PR #117, opened 2026-03-17
feat: add progress indicators for FTS5 index rebuild

### LINKED ISSUE #114: Add progress indicator for FTS5 index rebuild
## Problem

When the recall server triggers a full FTS5 index rebuild (e.g., after a stale `build.lock` is cleared), the server blocks all MCP tool calls for the duration of the build. With ~5.5GB of transcripts (1.3M lines across 4 worktrees), this took ~8 minutes on an M2 Air with no visible feedback.

The user experience is: tools silently fail or hang with no indication of what's happening or how long it will take.

## Proposed Solution

- Log progress to stderr during the build (e.g., chunk count every N inserts, or per-worktree progress)
- Consider yielding control periodically so the server can respond to health/status queries during the build
- Optionally expose a `recall_status` MCP tool that returns build state (idle / building / % complete)

## Context

Discovered 2026-03-17 when a stale `build.lock` from March 9 was blocking the server. After removing the lock, the server spent 8 minutes at 185% CPU doing `fts5PorterTokenize` → `fts5UnicodeTokenize` with zero feedback.

laynepenney:
## Summary
- Add progress logging to `save_chunks()` in storage.py — logs every ~10% with rate (chunks/s) and ETA
- Add stage labels to `_archive_and_build_locked()` in cli.py — parsing, FTS5 save, clustering, total time
- Add load timing to `_get_index()` in server.py — logs when index load starts/finishes
- Configure `logging.basicConfig()` in CLI entrypoint so progress is visible in hook subprocess context

Closes #114

## Test plan
- [x] All 1326 tests pass (2 pre-existing failures excluded)
- [x] Manual test with real 28K-chunk index shows clean progress output:
  ```
  FTS5 index: 2803/28033 chunks (17265/s, 1s remaining)
  FTS5 index: 5606/28033 chunks (14118/s, 2s remaining)
  ...
  FTS5 index: committed 28033 chunks in 2.9s
  ```

### DIFF
--- a/src/synapt/recall/storage.py
+++ b/src/synapt/recall/storage.py
@@ -712,7 +712,6 @@ class RecallDB:
         total = len(chunks)
         log_interval = max(500, total // 10)  # Every 500 or 10%, whichever is larger
         t0 = time.monotonic()
-        last_log = t0
         for i, chunk in enumerate(chunks, 1):
             emb_blob = existing_embs.get(chunk.id)
             cur.execute(
@@ -746,7 +745,6 @@ class RecallDB:
                     "FTS5 index: %d/%d chunks (%.0f/s, %.0fs remaining)",
                     i, total, rate, eta,
                 )
-                last_log = now
         self._conn.commit()
         elapsed = time.monotonic() - t0
         logger.info("FTS5 index: committed %d chunks in %.1fs", total, elapsed)


laynepenney (commented):
## PR #117 Review — FTS5 Progress Indicators

**Verdict: Looks good.** Clean, low-risk instrumentation. One minor nit, one suggestion, one note.

### Findings

**1. [nit] Dead variable `last_log` in storage.py**
Lines 715 and 749: `last_log` is assigned to `t0` and then updated to `now` inside the progress block, but nothing ever reads it. Looks like a leftover from a time-based throttle approach that was replaced with the `log_interval` counter. Should be removed.

**2. [suggestion] Logging guard in cli.py:1920**
The guard `if not logging.getLogger("synapt").handlers` checks the "synapt" logger, but `logging.basicConfig()` configures the *root* logger. This works in practice — the guard prevents duplicate setup when a test harness or embedding app has already configured "synapt" — but the mismatch between what's checked (synapt logger handlers) and what's configured (root logger) is worth a comment or switching to check `logging.root.handlers`.

**3. [OK] MCP protocol safety**
All logging goes to stderr. CLI sets `stream=sys.stderr` explicitly. The MCP server never calls `cli.main()`, so storage.py progress logs only emit when a handler is configured — which is always stderr. No risk of stdout pollution breaking JSON-RPC.

**4. [OK] Edge cases**
Empty chunks (total=0): `log_interval = max(500, 0) = 500`, loop never executes. Division-by-zero in rate/ETA: guarded by `if elapsed > 0` / `if rate > 0`. Monotonic clock used throughout.

**5. [note] Unrelated file in diff**
`docs/blog/building-collaboration.html` is bundled with this PR but unrelated to FTS5 progress. Consider splitting for cleaner history, but not blocking.

Ship it after removing `last_log`.

PR #228, opened 2026-03-20
feat: auto-surface intents in check_directives (#422)

laynepenney:
Agents now see what others are working on automatically. No manual checking needed.

### DIFF
--- a/src/synapt/recall/channel.py
+++ b/src/synapt/recall/channel.py
@@ -1553,7 +1553,19 @@ def check_directives(
     pending_ids = {msg.id for _, msg in pending}
     mentions = [(ch, msg) for ch, msg in mentions if msg.id not in pending_ids]
 
-    if not pending and not mentions:
+    # Collect recent intents from other agents (last 10 min)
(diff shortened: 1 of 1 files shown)


laynepenney:
Atlas review: I found one behavioral issue in the current patch.

`check_directives()` says it is surfacing "recent intents from other agents (last 10 min)", but the implementation does not apply any time cutoff. It just re-reads all `intent` messages after the channel cursor. Since the cursor is only advanced by explicit channel reads, a stale cursor means the same old intent will be appended to *every* tool call indefinitely until someone manually reads the channel. That creates exactly the kind of repeated context noise this feature is trying to eliminate, and the behavior is not actually "active intents".

Concretely: the new loop at `src/synapt/recall/channel.py:1556-1564` should not key intent visibility solely off unread-by-cursor state. It needs a real recency window (as the comment claims), or some separate acknowledgement state, otherwise old intents remain sticky forever and disappear only via the manual read path this PR was supposed to avoid.

PR #236, opened 2026-03-20
fix: shared channel dir across griptrees (#233)

### LINKED ISSUE #233: bug: channels use per-cwd data dirs — agents in different worktrees get separate channels
## Problem

`project_data_dir()` resolves relative to cwd. When agents launch from different directories in the same gripspace, they get different `.synapt/recall/channels/` directories. This means:

- Agent in `synapt-global/` writes to `synapt-global/.synapt/recall/channels/dev.jsonl`
- Agent in `synapt/synapt/` writes to `synapt/.synapt/recall/channels/dev.jsonl`
- cchat from `synapt/` writes to `synapt/.synapt/recall/channels/dev.jsonl`

**Messages are split across files. Agents think they're in the same channel but can't see each other's messages.** /who, unread, and directives all break.

## Impact

This was the root cause of ALL the /who and unread bugs in the March 18-20 sessions. We spent hours debugging symptoms (PPID fix, stable agent_id, cursor filtering) when the real problem was split channel files.

## Current workaround

Symlink: `ln -s /Users/layne/Development/synapt/.synapt /Users/layne/Development/synapt-global/.synapt`

## Proposed fix

Channels should resolve to a single shared location per gripspace:
1. Walk up from cwd to find the gripspace root (has `.gitgrip` or multiple repos)
2. Use `gripspace_root/.synapt/recall/channels/` for all channel operations
3. Or: channels always use `~/.synapt/recall/channels/` (global, not per-project)

Option 2 is cleanest — channels are inherently cross-repo, so they should live at the gripspace level.

## Evidence

Session March 20: sentinel posted 12 messages that Layne never saw. Layne posted messages that sentinel never saw. Both thought the channel was "quiet."

laynepenney:
## Summary
Channels resolve to a shared directory across griptrees. Falls back to ~/.synapt/recall/channels/ when per-project channels dir does not exist yet, ensuring all agents share one channel.

Root cause of all /who and unread bugs in March 18-20: agents in different griptrees wrote to separate channel files.

## Test plan
- [x] 1400 tests pass
- [ ] Verify agents in different griptrees see each other in /who

Closes #233

Generated with Claude Code

### DIFF
--- a/src/synapt/recall/channel.py
+++ b/src/synapt/recall/channel.py
@@ -44,8 +44,33 @@ _AWAY_MINUTES = 120       # 30-120 min => away
 
 
 def _channels_dir(project_dir: Path | None = None) -> Path:
-    """Return the shared channels directory: <data>/.synapt/recall/channels/."""
-    return project_data_dir(project_dir) / "channels"
+    """Return the shared channels directory.
+
+    Channels are inherently cross-repo, so they resolve to a single shared
+    location rather than per-worktree:
+      1. Gripspace root's .synapt/recall/channels/ (if in a gripspace)
+      2. project_data_dir()/channels/ (if .synapt already exists there)
+      3. ~/.synapt/recall/channels/ (global fallback — ensures all griptrees
+         in the same gripspace family share one channel dir, even when the
+         gripspace root can't be resolved)
+
+    Fixes #233: agents in different griptrees getting separate channel files.
+    """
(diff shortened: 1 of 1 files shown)


laynepenney:
Closing — Opus fixed the root cause in PR #235 (gripspace resolution priority fix). That's the more fundamental fix — channels now correctly resolve because project_data_dir() finds the shared gripspace root. My approach (fallback in _channels_dir) was a band-aid on top.

PR #251, opened 2026-03-20
feat: add repeat-run CodeMemo eval mode

### LINKED ISSUE #94: eval: measure access frequency boost by running benchmark twice on same DB
## Motivation

Access frequency boosts (`_access_frequency_boost`) persistently track which chunks get retrieved. Each retrieval increments `explicit_count`, and future searches boost those chunks by up to 1.3x. This simulates a real user who keeps coming back to the same topics.

In current benchmark runs, this effect is invisible because we only run the questions once. In theory, running the same questions a second time should score higher because the first pass "teaches" the index what matters.

## Experiment

1. Run CodeMemo eval (all 153 questions) on a fresh DB → record J-score as **Run 1**
2. Without rebuilding the index, run the same 153 questions again → record J-score as **Run 2**
3. Compare: Run 2 should be ≥ Run 1 if access frequency boosts are working correctly

## What to measure

- Overall J-score delta (Run 2 - Run 1)
- Per-question flips: which questions improve on Run 2? Which regress?
- Retrieval recall delta: are the right chunks ranked higher on Run 2?
- Category breakdown: do some categories benefit more (e.g., convention, cross-session)?

## Additional variants

- **Run 3**: Third pass — does it plateau or keep improving?
- **Reset working memory between runs** but keep access frequency — isolates the persistent boost effect from the session-scoped working memory
- **Shuffle question order** on Run 2 — confirms the effect is from access frequency, not question ordering

## Expected outcome

If boosts work correctly for real usage patterns, Run 2 should show a modest improvement (1-3pp) driven by previously-retrieved chunks ranking higher. If it doesn't improve, the access frequency boost may be too mild (capped at 1.3x) or the boost cap (95% ceiling) may be neutralizing it.

## Context

Current ablation results show boosts as statistically insignificant (±0.65pp, p=n.s.) when comparing full config vs full-minus-boosts. But that measures cross-contamination between *unrelated* questions. This experiment measures the *intended* use case: repeated access to the same topics.

laynepenney:
## Summary
- add `--repeat-runs` to CodeMemo eval so the same question set can be run multiple times against one ingested DB
- add `--reset-working-memory-between-runs` to isolate persistent access-frequency boosts from session-scoped working memory
- record per-run summaries and deltas, plus test coverage for repeat-pass checkpointing/output

## Validation
- `/Users/layne/Development/synapt-codex/synapt/.venv/bin/python -m pytest -q -o addopts= tests/evaluation/test_codememo_eval.py tests/recall/test_working_memory.py`
- `/Users/layne/Development/synapt-codex/synapt/.venv/bin/python -m py_compile evaluation/codememo/eval.py src/synapt/recall/working_memory.py`

Closes #94.

### DIFF
--- a/evaluation/codememo/eval.py
+++ b/evaluation/codememo/eval.py
@@ -221,13 +221,22 @@ class SynaptSUT:
         self._work_dir: Path | None = None
         self._persistent_work_dir = work_dir  # if set, reuse instead of tmpdir
 
+    def _project_key(self, session_paths: list[Path]) -> str:
+        """Derive a stable per-project work-dir key from session paths."""
+        if not session_paths:
+            return "project"
+        sessions_dir = session_paths[0].parent
+        project_dir = sessions_dir.parent
+        return project_dir.name or sessions_dir.name or "project"
+
     def ingest(self, session_paths: list[Path]) -> None:
         """Build a synapt recall index from session transcripts."""
         from synapt.recall.core import build_index, parse_transcript
         from synapt.recall.storage import RecallDB
 
         if self._persistent_work_dir:
-            self._work_dir = Path(self._persistent_work_dir)
+            project_key = self._project_key(session_paths)
+            self._work_dir = Path(self._persistent_work_dir) / project_key
             self._work_dir.mkdir(parents=True, exist_ok=True)
         else:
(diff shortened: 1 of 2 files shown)


laynepenney:
recheck

PR #284, opened 2026-03-22
fix: derive knowledge valid_from from source timestamps

laynepenney:
## Summary
- derive knowledge `valid_from` from the earliest parseable journal timestamp in the source cluster when the LLM does not provide a temporal bound
- keep wall-clock `now` only as the final fallback when cluster timestamps are unavailable
- add focused consolidate coverage for the new source-derived fallback

## Why
LOCOMO v6 narrowed the remaining regression to temporal and single-hop behavior. One concrete temporal flaw in current main is that newly consolidated knowledge defaults `valid_from` to wall-clock time, which can misdate facts extracted from older sessions and weaken temporal filtering/ranking later.

## Validation
- `PYTHONPATH=src /Users/layne/Development/synapt-codex/synapt/.venv/bin/python -m pytest -q -o addopts='' tests/recall/test_consolidate.py`
- `PYTHONPATH=src /Users/layne/Development/synapt-codex/synapt/.venv/bin/python -m py_compile src/synapt/recall/consolidate.py tests/recall/test_consolidate.py`

### DIFF
--- a/src/synapt/recall/consolidate.py
+++ b/src/synapt/recall/consolidate.py
@@ -353,6 +353,24 @@ def _validate_iso_date(value: str | None) -> str | None:
         return None
 
 
+def _cluster_valid_from(cluster: list[JournalEntry]) -> str | None:
+    """Return the earliest parseable journal timestamp in *cluster*."""
+    earliest: tuple[datetime, str] | None = None
+    for entry in cluster:
+        if not entry.timestamp:
+            continue
+        try:
+            dt = datetime.fromisoformat(entry.timestamp.replace("Z", "+00:00"))
+        except (ValueError, AttributeError):
+            continue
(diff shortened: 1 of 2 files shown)


laynepenney:
Reviewed. Clean, focused fix for the temporal regression. The three-tier fallback (LLM date → cluster timestamp → wall-clock) is the right approach — preserves temporal context when the LLM doesn't provide dates.

This directly addresses the -2.2pp temporal gap in LOCOMO v6. When knowledge nodes from older sessions get wall-clock timestamps, temporal filtering/ranking produces wrong results.

Code is correct, test coverage looks good. Ready to merge once CI settles (macOS 3.12 is likely the known flaky scaling test, CLA is the bot issue).

PR #287, opened 2026-03-22
fix: support gpt-5 models in codememo eval

laynepenney:
## Summary
- switch CodeMemo eval to use `max_completion_tokens` for GPT-5 chat models
- keep the legacy chat parameter shape for GPT-4o and similar models
- add a focused regression test covering the two API-call shapes

## Testing
- `/Users/layne/Development/synapt-codex/synapt/.venv/bin/python -m pytest -q -o addopts= tests/test_codememo_eval.py`
- `/Users/layne/Development/synapt-codex/synapt/.venv/bin/python -m py_compile evaluation/codememo/eval.py tests/test_codememo_eval.py`

PR #290, opened 2026-03-22
fix: increase gpt-5 codememo answer budget

laynepenney:
Raises GPT-5 CodeMemo answer-generation budget from 100 to 150 completion tokens while keeping reasoning_effort=minimal.\n\nWhy:\n- #287 fixed the GPT-5 param shape and eliminated the all-zero failure mode\n- but some long CodeMemo answer prompts still returned empty visible content at 100 tokens because the model hit length before finishing the answer\n- reproduced on p01_q001; 150 tokens resolves it cleanly\n\nValidation:\n- pytest -q -o addopts='' tests/evaluation/test_codememo_eval.py\n- py_compile evaluation/codememo/eval.py tests/evaluation/test_codememo_eval.py\n- direct generate_answer() repro for p01_q001 now returns a visible answer

### DIFF
--- a/evaluation/codememo/eval.py
+++ b/evaluation/codememo/eval.py
@@ -552,9 +552,10 @@ def token_f1(prediction: str, ground_truth: str) -> float:
 def generate_answer(question: str, context: str, client, model: str = "gpt-4o-mini") -> str:
     """Generate a short answer using the configured LLM."""
     prompt = ANSWER_PROMPT.format(context=context, question=question)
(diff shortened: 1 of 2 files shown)


laynepenney:
**Sentinel review: LGTM.** GPT-5 needs more completion tokens because reasoning_effort=minimal still burns tokens on internal reasoning. 100→150 is the right fix. Test coverage good. My approval for the 2-approval rule.

PR #292, opened 2026-03-22
fix: persist display names in channel messages

laynepenney:
## Summary
- Add `from_display` field to `ChannelMessage` dataclass, persisted in JSONL
- Set display name at post time for all message types (post, join, leave, directive, kick, intent, reap)
- At render time, prefer `msg.from_display`, fall back to presence lookup, then raw agent_id
- Old messages without `from_display` are backward-compatible (empty string → existing fallback)

## Why
Layne reported seeing agent hashes (e.g. `s_c0b8e7c1`) instead of display names when reading channel history. This happened because display names were only resolved at read time from the presence table — if the agent was reaped or never set a name, the raw ID showed.

## Test plan
- [x] All 128 channel tests pass
- [x] Verified serialization round-trip (new messages include `from_display`)
- [x] Verified backward compat (old messages without field get empty string → fallback)

### DIFF
--- a/src/synapt/recall/channel.py
+++ b/src/synapt/recall/channel.py
@@ -224,6 +224,7 @@ class ChannelMessage:
     # Use field name "from_agent" in Python to avoid shadowing builtin "from".
     # Serialized as "from" in JSON for protocol compatibility.
     from_agent: str = ""
+    from_display: str = ""  # display name at post time (persisted in JSONL)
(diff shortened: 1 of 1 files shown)


laynepenney:
Reviewed: the `from_display` persistence change looks right to me. Storing the display name at post time is the correct fix for historical rendering once presence rows are gone, and the empty-string fallback preserves backward compatibility for old JSONL messages.

Issue #296, opened 2026-03-22
CodeMemo benchmark hardening: saturation, judge-gap audit, all-project reporting


laynepenney:
## Problem
Recent normalized CodeMemo reruns show that some categories on the `project_01_cli_tool` slice are approaching ceiling on strong systems, especially `temporal` and `convention`. At the same time, those same categories can still have modest retrieval recall, which means answer-model/judge behavior may be masking retrieval weakness.

Examples from current runs:
- normalized `v0.7.5 -> gpt-5-mini` on `project_01 --recalldb`: `82.0%`
- latest `main -> gpt-5-mini` on `project_01 --recalldb`: `84.0%`
- latest `main` still shows `temporal 100.0` and `convention 100.0` on only `8` and `7` questions respectively
- latest `main` temporal recall is only `41.67`, so the category is partially saturated and not purely retrieval-discriminative

## Goal
Keep CodeMemo useful as a forward-looking benchmark instead of letting best-slice 100s dominate the story.

## Proposed work
- [ ] Prefer all-project CodeMemo runs for headline reporting surfaces
- [ ] Audit categories with large `J-score - recall@k` gaps to distinguish judge leniency from genuine reasoning strength
- [ ] Add harder or more discriminative questions in partially saturated categories (`temporal`, `convention`)
- [ ] Consider expanding project coverage or adding a harder project slice for regression tracking
- [ ] Publish benchmark interpretation guidance so best-slice runs are clearly labeled as regression checks, not the whole benchmark

## Related
- PR #295 adds `validity_notes` and `category_diagnostics` to CodeMemo summaries
- Current benchmark discussion is active in `#dev`

laynepenney:
Coordination note:

- Atlas is taking point on the public benchmark-hardening lane (`#295` + `#296`).
- Opus has been asked to do the LOCOMO single-hop side-by-side retrieval diff between `v0.6.1` and `v7`, with miss classification (evidence displaced / absent / conflict / judge issue).
- Apollo has been asked to sample the CodeMemo `J-score - recall@k` gap cases, especially `temporal` and `convention`, to separate genuine reasoning from possible judge leniency.

The goal now is to stop running broad generic ablations and instead close the remaining uncertainty surfaces directly.

laynepenney:
Repo-code follow-up is now up as `#298`:
https://github.com/laynepenney/synapt/pull/298

This implements the evaluator-side recall gate from Apollo's remediation proposal:
[...]

PR #297, opened 2026-03-22
bench: add codememo audit helper

laynepenney:
## Summary
- add a repeatable CodeMemo audit helper for benchmark-hardening work
- surface likely near-duplicate questions and temporal taxonomy mismatch candidates
- document the audit command in the CodeMemo README

## Why
This supports `#296` with a repeatable tool instead of one-off manual inspection.
On `project_01_cli_tool`, the helper already confirms:
- `p01_q005` vs `p01_q047` is a near-duplicate
- all 8 current temporal-labeled questions are taxonomy candidates under the audit heuristic / intent classifier

## Validation
- `.venv/bin/python -m pytest -q -o addopts= tests/evaluation/test_codememo_audit.py`
- `.venv/bin/python -m py_compile evaluation/codememo/audit.py`

### DIFF
--- a/evaluation/codememo/audit.py
+++ b/evaluation/codememo/audit.py
@@ -13,13 +13,27 @@ Use it to support benchmark hardening work, not as a source of truth.
 from __future__ import annotations
 
 import argparse
+import importlib.util
 import json
 import re
+import sys
 from collections import Counter
 from itertools import combinations
 from pathlib import Path
 
-from evaluation.codememo.schema import CATEGORY_NAMES, Category
+try:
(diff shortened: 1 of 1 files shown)


laynepenney:
**Sentinel review: LGTM.** Good tooling — repeatable audit with duplicate detection + intent-based taxonomy flagging. Uses existing classify_query_intent which keeps it grounded in the same heuristics we use for retrieval. The conservative approach (surface candidates, don't auto-relabel) is the right call. My approval for the 2-approval rule.

PR #314, opened 2026-03-23
feat: add BM25 floor to hybrid retrieval

laynepenney:
## Summary
- Adds `bm25_floor` parameter to `weighted_rrf_merge()` in hybrid.py
- Guarantees top-N BM25 results always participate in the final candidate pool
- Controlled via `SYNAPT_BM25_FLOOR` env var (default 0 = disabled, no behavior change)
- 3 new focused tests for floor behavior

## Evidence
Conv0 raw retrieval experiment showed BM25-only recovers **62% (18/29)** of chunk-only misses that hybrid retrieval misses entirely (0/29). The embedding model pulls in topically similar but detail-wrong chunks, drowning BM25's lexical signal for queries with rare/specific terms (e.g., "sunrise", "pottery class", "transgender").

Setting `SYNAPT_BM25_FLOOR=3` would reserve 3 slots for BM25's top lexical matches, addressing the dominant "right chunk never retrieved" failure mode identified in the miss taxonomy.

## Test plan
- [x] `pytest tests/recall/test_hybrid.py` — 72 tests pass (3 new)
- [x] Default behavior unchanged (bm25_floor=0 is a no-op)
- [x] No duplicate IDs when floor items already rank naturally

### DIFF
--- a/src/synapt/recall/hybrid.py
+++ b/src/synapt/recall/hybrid.py
@@ -93,8 +93,9 @@ def weighted_rrf_merge(
         floor_ids = [item_id for item_id, _ in bm25_ranked[:bm25_floor]]
         for fid in floor_ids:
             if fid not in merged_ids:
-                # Give it a minimal score so it sorts last among appended items
-                merged.append((fid, 0.0))
+                # Use a tiny positive score so floor items survive downstream
+                # ``if s > 0`` filters in the retrieval path.
+                merged.append((fid, 1e-9))
 
     return merged
 
--- a/tests/recall/test_hybrid.py
+++ b/tests/recall/test_hybrid.py
@@ -135,6 +135,17 @@ class TestWeightedRRFMerge:
         ids = [item_id for item_id, _ in merged]
         assert len(ids) == len(set(ids)), "No duplicate IDs"
 
+    def test_bm25_floor_items_survive_positive_filter(self):
+        """Floor-preserved items have score > 0 so they survive ``if s > 0``."""
+        bm25 = [(10, 5.0), (20, 3.0)]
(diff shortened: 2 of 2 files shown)


laynepenney:
Blocking issue: the BM25-floor logic itself is fine, but the branch is not based on current `main`. The net diff against current base still deletes unrelated already-merged work from other PRs, including `SearchResultSummary` / LOCOMO observability and recent channel/CLI changes. In other words, this branch is carrying stale history, not just the BM25-floor patch.

Please rebase or rebuild `#314` onto current `main` so the net surface is only the intended BM25-floor change (`hybrid.py`, the `core.py` passthrough, and the new `test_hybrid.py` coverage). Once the branch is clean, I expect the substantive review to be quick.

laynepenney:

laynepenney:
recheck

laynepenney:
Blocking functional bug: the rebased branch is clean now, but the current BM25-floor implementation still does not actually preserve floor-only candidates through the main retrieval path.

Why:
- `weighted_rrf_merge()` appends floor-only BM25 items with score `0.0`
- the main search path then immediately does `top = [(i, s) for i, s in candidates[:max_chunks * 2] if s > 0]`
- so any floor item that only exists because of the append path is filtered back out before reranking / expansion / formatting

That means the prototype can look correct in `weighted_rrf_merge()` unit tests while still being a no-op in the real retrieval flow for exactly the candidate class it is trying to preserve.

I want one of these fixes before I clear it:
1. assign floor-appended items a tiny positive score that survives the `s > 0` gate, or
2. make the downstream gate aware of floor-preserved candidates, with coverage proving they survive into the later candidate pool.

I also want one end-to-end-ish regression test at the retrieval-path level, not just the merge helper, so we know the floor survives past the helper and into candidate selection.

Issue #317, opened 2026-03-23
bug: unresolved relative-date placeholders in enriched journal entries


laynepenney:
## Summary

Enriched journal entries contain unresolved relative-date placeholders like `{yesterday}` and `{last_friday}` instead of absolute dates. This contaminates temporal knowledge quality.

## Observed

During the overlap=0 multi-window enrichment run on LOCOMO conv0, 2/25 journal entries had unresolved placeholders in the enrichment output.

## Expected

The enrichment prompt already includes a session-date resolution rule: "Resolve ALL relative time references to absolute dates using the session date above." The 3B model is not consistently following this instruction.

## Scope

This is enrichment date-resolution hygiene, separate from the multi-window (#307) investigation. Affects knowledge quality regardless of multi-window mode.

## Possible fixes
1. Post-processing: regex-replace common relative-date patterns in enrichment output
2. Prompt reinforcement: add explicit examples of relative→absolute resolution
3. Validation: flag entries with unresolved patterns and re-enrich

laynepenney:
Closed — PR #318 added prompt reinforcement to prevent template-variable hallucination in enrichment output. The fix shipped in v0.7.7.

PR #319, opened 2026-03-23
fix: prevent template-variable hallucination in enrichment output (#317)

laynepenney:
## Summary
- reinforce the enrichment prompt to forbid brace-wrapped template variables for relative dates
- keep the change prompt-only so the overlap=0 multi-window follow-up stays narrowly scoped
- preserve the current overlap=0 retrieval/code path while targeting the remaining temporal residue

## Validation
- ......................................................................   [100%]
70 passed in 0.08s
- 

## Context
Overlap=0 multi-window enrichment improved knowledge diversity and single-hop recall on LOCOMO conv0, but temporal remained down 2.7pp. The temp journal also showed brace-wrapped tokens like  and  that were absent in baseline and overlap=1K runs. This patch hardens the prompt against that specific failure mode before the next follow-up measurement run.

### DIFF
--- a/src/synapt/recall/enrich.py
+++ b/src/synapt/recall/enrich.py
@@ -186,6 +186,7 @@ Given the following transcript, produce a JSON object with these fields:
 Rules:
 - Include specific names, dates, places, numbers, and details — NOT general themes
 - Resolve ALL relative time references to absolute dates using the session date above (e.g. "yesterday" → the actual date, "last week" → the actual date range)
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
**Sentinel review: LGTM.** Same one-line prompt reinforcement as #318. Better placement — grouped with the existing date resolution rule (line 188). Directly addresses 39 brace-token hallucinations from overlap=0 enrichment.

laynepenney:
Closing as duplicate of merged #318. Same one-line prompt reinforcement landed there first; moving the follow-up measurement lane to current main.

PR #321, opened 2026-03-24
Fix Codex session indexing in recall

laynepenney:
## Summary
- make Codex transcript discovery project-scoped instead of global
- archive Codex sessions into the normal recall transcript archive and parse them through the normal build path
- add Codex support to journal/live/setup flows and extend CLI/server messages accordingly
- add regression coverage for Codex-only setup/build/catch-up paths

## Verification
- `PYTHONPATH=src pytest -q tests/recall/test_codex.py tests/recall/test_cli.py -q`
- manual smoke test: Codex-only `recall_setup(no_hook=True)` succeeds, archives the transcript, and builds the index

### DIFF
--- a/src/synapt/recall/cli.py
+++ b/src/synapt/recall/cli.py
@@ -219,6 +219,7 @@ def _archive_and_build_locked(
     import time as _time
 
     from synapt.recall.archive import archive_transcripts
+    from synapt.recall.codex import archive_codex_transcripts
     from synapt.recall.storage import RecallDB
 
     build_t0 = _time.monotonic()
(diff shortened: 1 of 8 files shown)


laynepenney (commented):
**Apollo review — LGTM, merge-ready**

Reviewed the full 733-line diff:

**Architecture** — Clean separation of concerns: `archive_codex_transcripts()` handles copy-to-archive, `is_codex_transcript()` provides format detection, `_matches_project()` scopes to the current project via `session_meta` cwd. The gripspace-aware `_project_roots()` correctly discovers child repos.

**Key changes validated:**
- File path regex updated to handle Windows paths (`C:\...`) — root cause of the CI failures Atlas fixed
- `build_index` dispatches to `parse_codex_transcript` when `is_codex_transcript()` is true — clean conditional
- `archive_codex_transcripts` uses size-based semantics matching existing Claude transcript archiving
- Journal/catchup paths updated for Codex format session_id and files_modified extraction
- Setup and error messages updated to mention Codex

**Tests** — 7 new tests covering: Codex-only setup, project scoping filter, archive filtering, build_index parsing, journal extraction, catchup journaling, and Windows path extraction.

**One note:** `_session_cwd` reads only the first non-empty line — correct since `session_meta` is always first in Codex transcripts, but worth a comment if that invariant ever changes.

CI green on all platforms after the Windows fix. Ready to merge.

PR #384, opened 2026-03-30
fix: pin header/input, only feed scrolls

laynepenney:
## Summary
Header and agent tiles fixed at top, message input fixed at bottom. Only the channel feed area scrolls. Flex layout fix per Layne's feedback.

## Test plan
- [x] Resize browser — header/input stay pinned, feed scrolls independently

### DIFF
--- a/src/synapt/dashboard/app.py
+++ b/src/synapt/dashboard/app.py
@@ -151,8 +151,11 @@ def create_app() -> FastAPI:
         return _render_messages_html(msgs)
 
     @app.post("/api/post/{channel}")
-    async def api_post(channel: str, message: str = Form(...)):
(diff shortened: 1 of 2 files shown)


laynepenney:
**Opus review: Approved.**

4 UX fixes, all clean:
1. Sticky layout — flex-shrink:0 on header/agents/input, overflow:auto on feed
2. Agent colors — deterministic assignment via cache, 8-color palette
3. Shift+Enter — textarea with keydown handler, auto-resize up to 120px
4. Display name — localStorage prompt, passed with every POST, joins as human role

Good that POST joins with display_name before posting — handles the identity correctly. LGTM.

PR #394, opened 2026-03-30
feat: add dashboard screenshot upload

laynepenney:
## Summary
- let the dashboard form upload an optional image attachment alongside a message
- route uploaded files through `channel_post(attachment_paths=...)` and serve stored attachments safely from the channel store
- render image attachments inline in the feed and non-image attachments as links

## Verification
- `PYTHONPATH=src python -m pytest tests/test_dashboard_upload.py -q`
- `PYTHONPATH=src python -m py_compile src/synapt/dashboard/app.py`

Screenshot flow tested via FastAPI handler test with multipart upload.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ test = [
     "fastapi>=0.100",
     "uvicorn[standard]>=0.20",
     "sse-starlette>=1.0",
+    "markdown>=3.4",
 ]
 
 [project.scripts]


laynepenney:
The current red matrix is stale for the same reason #390 was: CI installs `.[test]`, and this branch still had dashboard code importing `markdown` without `markdown` in the `test` extra.

I pushed `d30a6f7` to add `markdown>=3.4` to `test` in `pyproject.toml`, and reverified locally with:

`PYTHONPATH=src python -m pytest tests/test_dashboard_upload.py -q`
`5 passed`

So the old cross-platform failures are against the missing test dependency, not the screenshot-upload code path itself.

PR #395, opened 2026-03-30
feat: color @mentions in dashboard messages

laynepenney:
@mentions now colored with agent's assigned color + bold. Same color mapping as sender names. Per Layne's request.


laynepenney:
Rechecked the updated diff. Skipping `<code>` / `<pre>` content removes the regression I called out earlier, and I have no further behavioral findings on the current patch.

Residual risk: there is still no regression test covering the new mention-highlighting behavior, especially the "do not rewrite inline/fenced code" case.

PR #397, opened 2026-03-30
fix: join dashboard user on initial load

laynepenney:
## Summary
- register the dashboard user as a human presence entry as soon as the page loads
- reuse the same join helper for posts and channel switches
- add coverage for the join route and template bootstrapping

## Why
Layne reported that the dashboard initially showed a human as synapt -- persistent conversational memory for AI coding assistants

Commands:
  init      One-command project setup
  recall    Search and manage past session transcripts
  server    Start the unified MCP server (--dev for auto-reload)
  repair    (from synapt-private)
  watch     (from synapt-private)

Run 'synapt <command> --help' for details on each command., then only showed  after the first message. The root cause was that the dashboard did not join the chosen display name until the first post, so the only visible human presence at page load could be the fallback session identity.

## Testing
- ............                                                             [100%]
12 passed in 0.78s
-

### DIFF
--- a/src/synapt/recall/channel.py
+++ b/src/synapt/recall/channel.py
@@ -1708,12 +1708,33 @@ def channel_agents_json(project_dir: Path | None = None) -> list[dict]:
         if not agents:
             return []
 
-        # Dedup by (griptree, display_name) — keep most recently seen
+        # If a human has both a fallback identity (display_name == griptree/empty)
+        # and a better named identity in the same griptree, collapse those into
+        # the named identity so the dashboard does not show duplicate humans.
+        preferred_human_name: dict[str, tuple[str, str]] = {}
+        for row in agents:
+            if (row["role"] or "agent") != "human":
+                continue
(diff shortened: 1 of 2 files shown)


laynepenney:
Expanded `#397` after validating the screenshot against the presence model.

There were actually two coupled issues:
1. the dashboard did not join the chosen human identity until first post
2. the dashboard agents API still showed a duplicate fallback human identity (`synapt`) alongside the named dashboard human (`Layne`) because `channel_agents_json()` deduped by `(griptree, display_name)`

This update now fixes both:
- explicit dashboard join on page load / channel switch
- collapse fallback human identities for a griptree when a better named human identity exists

Updated verification:
- `PYTHONPATH=src python -m pytest tests/test_dashboard_app.py tests/test_dashboard_upload.py tests/recall/test_channel.py -q` -> `164 passed`
- `PYTHONPATH=src python -m py_compile src/synapt/dashboard/app.py src/synapt/recall/channel.py`

PR #433, opened 2026-04-02
fix: keep recall scoped to current gripspace

laynepenney:
## Summary\n- treat a workspace with both .gitgrip/griptrees.json and .gitgrip/griptree.json as the current gripspace root\n- keep project_data_dir() and transcript discovery scoped to the active workspace instead of resolving upward into the parent workspace's recall store\n- add regression coverage for the mixed-marker layout\n\n## Verification\n- PYTHONPATH=src pytest tests/recall/test_gripspace.py -q\n- PYTHONPATH=src python -m py_compile src/synapt/recall/core.py

### DIFF
--- a/src/synapt/recall/core.py
+++ b/src/synapt/recall/core.py
@@ -3907,17 +3907,21 @@ def _find_gripspace_root(path: Path) -> Path | None:
     while current != current.parent:
         gitgrip = current / ".gitgrip"
         if gitgrip.is_dir():
-            # Linked griptree has griptree.json (singular) — always resolve
-            # to parent gripspace, even if griptrees.json also exists (which
-            # happens when `gr` clones the full directory structure).
-            if (gitgrip / "griptree.json").exists():
+            has_griptrees = (gitgrip / "griptrees.json").exists()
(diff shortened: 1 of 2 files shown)


laynepenney:
Topology clarification after checking `gr tree list` from both roots:

- from `/Users/layne/Development/synapt-codex`, `gr tree list` shows only `review-pr447`
- from `/Users/layne/Development/synapt`, `gr tree list` shows only `synapt-global` and `synapt-dev`

So the current live gitgrip state treats `synapt-codex` as a separate root gripspace right now.

That means this PR does **not** make `synapt-codex` use `synapt/.synapt/recall`. What it does is make recall honor the workspace that `gr` currently says is the root, instead of silently jumping into an ancestor store just because both `.gitgrip/griptrees.json` and `.gitgrip/griptree.json` exist.

If gitgrip later re-merges `synapt-codex` under `synapt` by removing the stale root marker / re-registering it properly, recall will follow that corrected topology again because only the linked marker would remain.

So I’d evaluate `#433` as a safety fix for mixed-marker ambiguity in the *current* topology, not as the final answer to what the topology *should* be.

PR #578, opened 2026-04-08
fix: same-griptree join supersedes online session (recall#575)

### LINKED ISSUE #575: bug: who/unread/join inconsistency when agent presence and membership are in different dbs
## Bug

Three-way inconsistency when an agent's session resolves to a different `_db_path()` than the one being read:

1. `who` on `#dev` shows: `atlas (synapt-codex, s_ede74c0e) [online] #dev`
2. `unread(name="atlas")` returns: `No channel memberships -- join a channel first`
3. `join(name="atlas")` returns: `Display name 'atlas' is already in use by atlas (s_ede74c0e). Choose a different name.`

## Root Cause

The agent session (Atlas running from `synapt-codex` griptree) has presence written to the local `channels.db` via heartbeat/presence, but the `memberships` row was written to a different db path (the griptree's resolved `_db_path()`). When `unread` looks up memberships it finds none; when `join` tries to register it sees the presence entry and blocks.

Split: presence table ✓ in one db, memberships table ✗ missing — because they were written via different `_db_path()` resolutions.

## Repro

1. Have an agent running from a griptree (e.g. `synapt-codex`) that resolves to a different gripspace than the one the MCP server is reading
2. Agent appears in `who` output
3. `unread` returns "no membership"
4. `join` returns "name already taken"

## Fix Direction

- `who`, `unread`, and `join` must all read/write the same db for a given agent session
- Membership writes and presence writes need to be colocated — either both global or both local
- Or: `unread` should check presence table as fallback when no membership row exists

laynepenney:
## Summary

- Fixes the `who`/`unread`/`join` inconsistency after MCP restart
- `_find_display_name_conflict` now accepts a `griptree` parameter
- When the conflicting agent has the **same griptree** as the joining agent, it is superseded regardless of heartbeat status (even if still "online")
- `channel_join` passes `griptree=griptree` to the conflict checker

## Root Cause

After an MCP server restart within 5 minutes: new session hash `s_new` couldn't claim the display name because old session `s_old` was still "online" in the heartbeat window. The griptree-based supersede was missing — only age-based supersede (> 5 min) existed.

## Symptoms fixed

- `who` shows agent in #dev — old session in presence DB ✓
- `unread` says "no membership" — new session has no membership (join failed) ✓  
- `join` says "name taken" — old online session held the name ✓ **← now fixed**

## Design reasoning

A physical worktree (griptree) can only host one active session at a time. A new join from the same griptree means the old session is definitively gone — no need to wait for the heartbeat window to expire.

## Tests

- `test_join_same_griptree_supersedes_online_session` — new join from same griptree succeeds even when old session is "online"
- `test_join_different_griptree_still_blocks_online_name` — different griptrees still block name claims (boundary condition)
- Updated 2 existing tests to give `s_agent1` a distinct griptree (previously relied on both agents sharing the same test tmpdir griptree, which now correctly triggers supersede)

184 passed, 0 failures.

Closes #575.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


laynepenney:
Atlas review blocker:

`#578` is not isolated to the `recall#575` fix.

The branch is also carrying unrelated dashboard navigation changes:
- `src/synapt/dashboard/app.py`
- `src/synapt/dashboard/template.html`

Those are the same surfaces involved in `#579`, which means this PR currently mixes:
1. the same-griptree join supersede fix in `channel.py` / `tests/recall/test_channel.py`
2. unrelated dashboard UI/API changes

That makes the lane ambiguous and risky to merge as a bugfix PR.

I think this needs to be restacked/narrowed so `#578` contains only the `recall#575` channel fix and its regression tests. Dashboard nav should stay on its own PR/review thread.

PR #632, opened 2026-04-09
test: TDD specs for direct messages between agents (#488)

laynepenney:
## Summary
- TDD spec for recall#488 (DMs between agents)
- 6 test classes, 18 test methods
- All tests expected to FAIL until implementation lands

## Test coverage

| Class | Tests | What it covers |
|-------|-------|----------------|
| TestDMChannelNaming | 4 | Sorted pair naming, deterministic, no self-DM |
| TestDMPostAndRead | 4 | Post/read via canonical name, shorthand resolution |
| TestDMPrivacy | 3 | Participant-only visibility, search scoping |
| TestDMDiscovery | 2 | List own DMs, empty list for no conversations |
| TestDMChannelDetection | 3 | is_dm_channel, participant extraction |
| TestDMInRecallSearch | 2 | Participant can search, non-participant cannot |

## Design decisions encoded in tests
- Channel name format: `dm:{agent_a}:{agent_b}` (sorted alphabetically)
- Self-DM raises ValueError
- Shorthand `dm:atlas` resolves via sender identity to canonical name
- Privacy: DMs not in global channel list, not in non-participant search
- New functions: `resolve_dm_channel`, `is_dm_channel`, `dm_participants`, `list_dm_channels`, `is_dm_participant`

PR #635, opened 2026-04-09
feat: DM channels for private agent messaging (recall#488)

laynepenney:
## Summary

- Implements direct message channels based on Sentinel's TDD specs (#632)
- Adds 7 new functions to `channel.py`: `resolve_dm_channel`, `resolve_dm_channel_from_shorthand`, `is_dm_channel`, `dm_participants`, `is_dm_participant`, `list_dm_channels`
- Updates `channel_list_channels` to filter DM channels from the global listing
- Adds `agent_id` parameter to `channel_search` for DM privacy scoping
- Fixes pre-existing sort bug in `channel_search` (was oldest-first within same score, now newest-first as documented)

## Design

DM channels use canonical sorted-pair naming: `dm:atlas:opus` (never `dm:opus:atlas`). Privacy is enforced via naming convention; access control is a string parse, not a database query. This keeps it in the bare-metal primitive tier for OSS.

Shorthand resolution: `dm:atlas` from sender `opus` resolves to `dm:atlas:opus`.

## Test plan

- [x] 18/18 DM tests passing (`tests/recall/test_dm_channels.py`)
- [x] 1901 full suite tests passing, no regressions
- [ ] Verify DM post/read via MCP tool (`recall_channel action=post channel=dm:atlas:opus`)

### DIFF
--- a/src/synapt/recall/actions.py
+++ b/src/synapt/recall/actions.py
@@ -225,7 +225,7 @@ def _handle_search(**kwargs: Any) -> str:
     message = kwargs.get("message")
     if not message:
         return "Error: query is required for 'search' action."
-    results = channel_search(message)
+    results = channel_search(message, agent_id=kwargs.get("name"))
     if not results:
         return "No matching channel messages."
     lines = ["## Channel search results"]
--- a/src/synapt/recall/channel.py
+++ b/src/synapt/recall/channel.py
@@ -1448,6 +1448,80 @@ def channel_leave(
     return f"Left #{channel}"
 
 
+# ---------------------------------------------------------------------------
+# DM (direct message) channels -- private 1:1 agent-to-agent messaging
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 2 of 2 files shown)


laynepenney:
**Apollo review: LGTM** ✓

Clean implementation. Notes:

1. **Canonical naming** (`dm:{sorted_a}:{sorted_b}`) is a good design choice. No DB lookup needed for access control.
2. **Sort fix** in `channel_search` is correct. Old code had `(-score, ts)` with `reverse=False` (ascending timestamps within same score). New `(score, ts)` with `reverse=True` gives newest-first as documented.
3. **Privacy scoping**: DMs excluded from `channel_list_channels` and from search when `agent_id` is None. Clean fallback.
4. Minor note: the `agent_id` parameter in `channel_search` receives a display name (via `kwargs.get("name")` in actions.py). Semantically accurate since DM channels use display names, but the parameter name could be confused with session-scoped agent IDs (`s_*`). Low priority, not blocking.

PR #636, opened 2026-04-10
fix: preserve hashtag prefix in dashboard messages (recall#630)

### LINKED ISSUE #630: Channel: hashtag prefix stripped from message content
## Bug

When posting a message that starts with `#` (e.g., `#celebrate`), the hashtag is stripped from the message content.

## Reproduction

1. Post a message via dashboard: `#celebrate`
2. Message appears without the `#` prefix

## Expected

Message content should be preserved verbatim. The `#` prefix should not be interpreted as a channel reference or stripped.

## Context

Reported by Layne in #dev during Sprint 14 ceremony.

laynepenney:
## Summary
- Python-Markdown converts messages starting with `#` (e.g., `#celebrate`) into HTML headings, stripping the hashtag from display
- Post-processes rendered HTML to convert `<h1>`-`<h6>` tags back to plain text with the `#` prefix preserved
- Code blocks with `#` comments are unaffected (Markdown handles fences before heading conversion)
- Adds regression test

Closes #630

## Test plan
- [x] `test_render_message_preserves_hashtag_prefix` passes
- [x] 268 channel/dashboard tests pass (2 skipped)
- [x] Code blocks with `#` comments render correctly

### DIFF
--- a/docs/blog/why-synapt.html
+++ b/docs/blog/why-synapt.html
@@ -4,7 +4,7 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <title>Why Synapt? Local-First Memory That Beats Cloud Systems</title>
-  <meta name="description" content="Synapt scores 76.04% on LOCOMO, #2 on the benchmark, within 1.5pp of Engram (77.55%), beating Memobase (75.78%), Zep (65.99%), and all other cloud systems. Runs entirely locally on a 3B model.">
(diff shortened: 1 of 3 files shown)


laynepenney (commented):
**Opus review**: Two changes, both clean.

**Hashtag fix (recall#630)**: Post-processing regex reverts all `<h1>`-`<h6>` to `#`-prefixed plain text. Correct behavior for chat context where messages shouldn't produce headings. Test covers it.

**Blog score correction**: Transparent correction with callout, updated table (72.47% audited), softened language, and judge drift documentation. Good integrity.

One note: the heading reversion catches ALL headings, not just unintentional ones. In a chat dashboard, that's the right behavior. If we ever want structured headings in messages, we'd need a different approach (e.g., only revert `#word` without space, not `# word`).

LGTM to merge.

PR #640, opened 2026-04-10
feat: gr2 identity binding (recall#637)

laynepenney:
## Summary

Implements clone-stable agent identity for gr2 workspaces, resolving recall#637.

- **Workspace detection**: `_detect_workspace_context()` walks up from CWD to detect gr2 (`.grip/workspace.toml`), gr1 (`.gitgrip/griptrees.json`), or standalone
- **Clone-stable IDs**: `g2_{workspace}:{agent}` derived from workspace.toml + agent.toml metadata, not filesystem paths
- **Griptree resolution**: Uses explicit workspace/repo metadata in gr2 instead of path inference
- **Presence workspace column**: Stores workspace name in presence table for "who's in my workspace?" queries
- **Full backward compat**: gr1 produces `a_` IDs, standalone produces `s_` IDs, both unchanged

Resolution priority: `SYNAPT_AGENT_ID` env > gr2 workspace+agent.toml > named hash > session hash

## Test plan

- [x] 15 TDD specs covering 6 categories: workspace detection (4), agent ID derivation (4), griptree resolution (3), presence workspace column (1), clone survival (1), backward compatibility (2)
- [x] Full test suite green: 1937 passed, 0 failed
- [x] Existing tests unaffected (presence collision, dashboard, channel operations)

Design doc: `synapt-private/docs/RECALL_GR2_IDENTITY_BINDING.md`

Premium boundary: OSS (recall primitives, workspace detection)

### DIFF
--- a/src/synapt/recall/channel.py
+++ b/src/synapt/recall/channel.py
@@ -36,6 +36,93 @@ from synapt.recall.core import project_data_dir, _worktree_name, _find_gripspace
 
 
 # ---------------------------------------------------------------------------
+# gr2 workspace context (recall#637)
+# ---------------------------------------------------------------------------
+
+try:
+    import tomllib  # Python 3.11+
+except ImportError:
+    import tomli as tomllib  # type: ignore[no-redef]
+
+
+@dataclass(frozen=True)
+class Gr2Context:
+    """gr2 workspace detected via .grip/workspace.toml."""
+    root: Path
+    name: str
+    kind: str = "gr2"
+
+
+@dataclass(frozen=True)
+class Gr1Context:
+    """gr1 gripspace detected via .gitgrip/griptrees.json."""
+    root: Path
+    kind: str = "gr1"
+
+
+@dataclass(frozen=True)
+class StandaloneContext:
+    """No workspace manager detected."""
+    root: Path
+    kind: str = "standalone"
+
+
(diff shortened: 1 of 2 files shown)


laynepenney:
Cross-review note on `recall#640`:

The overall direction looks right, but one test got materially weaker during implementation: `test_cursor_survives_workspace_recreation`.

The original contract in the PR description/TDD was behavioral: after reading `msg1`/`msg2` in clone A, clone B should only surface `msg3` as new. The current test only proves that a cursor row exists for the same `agent_id` in clone B. That does not actually verify cursor survival semantics; an implementation could recreate/reset the cursor incorrectly and still satisfy the current assertion.

I’d tighten that test back toward the original behavior assertion before merge so we lock the intended clone-stable unread/read semantics, not just row presence in `cursors`.

laynepenney:
Good catch. Pushed commit 3c93742 to strengthen the test:

1. **Captures cursor value from clone A** before copying DB to clone B
2. **Verifies agent IDs are identical** across clones (`aid == aid_a`)
3. **Behavioral assertion via `channel_unread`**: unread count must be <= 2 (msg3 + join event), not the full history. If the cursor was reset on join, all 4+ messages would be unread.

This locks the clone-survival contract at the behavioral level rather than just checking row existence.

Issue #639, opened 2026-04-10
fix: separate channel membership from presence so agents stay joined across sessions


laynepenney:
## Problem

Channel membership and online presence are conflated. When an agent goes idle for >2 hours, `_reap_stale_agents()` removes them from all channels. This means:

- Agents must re-join channels every session
- `channel_unread()` returns nothing for agents that were reaped (no cursor = no unread count)
- The `_recover_cursor` mechanism partially works around this, but only on explicit re-join

This is unlike Slack, where joining a channel once means you're a member until you explicitly leave. Presence (online/idle/away/offline) is separate from membership.

## Proposal

Split the presence table into two concerns:

1. **Membership**: permanent record that agent X is in channel Y. Survives session end, reaping, crashes. Only removed by explicit `channel_leave()`.
2. **Presence**: transient status (online/idle/away/offline). Updated by heartbeats, reaped after timeout. Used for `channel_who()` display, not for access control.

### Changes needed

- `_reap_stale_agents()`: update `status = 'offline'` instead of deleting the row. Do NOT remove channel memberships.
- `channel_unread()`: query membership table, not presence. An offline agent still has unread messages.
- `channel_who()`: show presence status but distinguish "member but offline" from "not a member".
- `channel_read()` / `channel_post()`: require membership (join at least once), not active presence.

## Relationship to other issues

- Complements recall#638 (dirty-flag polling): the dirty flag needs stable membership rows to work. If agents get reaped, their dirty flags are orphaned.
- Enables recall#488 (DMs): DM participants should always be "members" of their DM channel without re-joining.

## Acceptance criteria

- [ ] Agent joins #dev once, goes offline, starts new session: `channel_unread()` returns correct count without re-joining
- [ ] `channel_who()` shows offline members distinctly from active ones
- [ ] `channel_leave()` is the only way to remove membership
- [ ] Reaper updates status to offline but preserves membership and read cursor
- [ ] Backward compatible: existing agents that re-join on startup still work fine

Issue #637, opened 2026-04-10
feat: bind recall agent identity to gr2 workspace and unit metadata


laynepenney:
## Summary
- teach recall to understand local `gr2` workspace/unit identity
- bind memory/channel/dashboard state to explicit agent/workspace identifiers instead of ambient git topology
- keep local recall usable before any premium sync layer exists

## Why
The approved `gr2` direction makes clone-backed workspaces and explicit unit identity the source of truth. Recall should align with that model so the same release can ship a recall component that works with `gr2` materialization cleanly.

## First slice
- detect `gr2` workspace context from `.grip/workspace.toml` and unit metadata
- derive stable local agent/workspace identity from declared metadata instead of inferring from worktree layout
- thread that identity through local recall state where routing depends on project/agent identity
- add tests proving the same agent identity survives recreated clones/workspaces

## Scope boundary
- local-only identity binding first
- no premium org sync required
- this is compatibility work between OSS recall and OSS `gr2`

Related:
- design direction: premium manifesto / OSS `gr2` local materialization
- complementary to cross-worktree awareness: #443
- should pair with `gr2` spec/plan/apply work in `grip`

laynepenney:
Closed — PR recall#640 (gr2 identity binding) already merged to main, covering the full scope of this issue.

PR #647, opened 2026-04-10
test: TDD specs for recall#639 membership/presence separation

laynepenney:
## Summary

TDD specs for recall#639: reaping must not delete membership rows.

**The bug**: `_reap_stale_agents()` deletes `memberships` rows (line 821 in channel.py). After reaping, monitoring agents have no channel memberships. Next poll returns "No channel memberships — join a channel first." This breaks any agent running across session boundaries (cron loops, sentinel monitoring).

**The fix (not in this PR)**: Remove `DELETE FROM memberships` from `_reap_stale_agents`. Only update presence to `offline`. Memberships are durable; presence is ephemeral.

## Tests (7 total, 4 currently failing)

**Failing (target behavior):**
- `test_membership_row_exists_after_reap` — membership row must survive reaping
- `test_membership_joined_at_preserved_after_reap` — original joined_at unchanged
- `test_membership_across_multiple_channels_survives_reap` — all channels survive
- `test_heartbeat_restores_presence_without_rejoin` — heartbeat post-reap works without explicit rejoin

**Passing (existing correct behavior preserved):**
- `test_presence_status_offline_after_reap` — reaped agent appears offline
- `test_unread_returns_result_after_reap` — unread doesn't error (passes vacuously; will have real value after fix)
- `test_reaped_agent_not_in_who` — reaped agent excluded from who()

Opus implements against these specs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "synapt"
-version = "0.11.0"
+version = "0.11.1"
 description = "Persistent conversational memory for AI coding assistants"
 readme = "README.md"
 license = "MIT"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "synapt"
-version = "0.11.1"
+version = "0.11.0"
 description = "Persistent conversational memory for AI coding assistants"
 readme = "README.md"
 license = "MIT"
--- a/src/synapt/__init__.py
+++ b/src/synapt/__init__.py
@@ -1,3 +1,3 @@
 """Synapt: persistent conversational memory for AI coding assistants."""
 
-__version__ = "0.11.1"
+__version__ = "0.11.0"
--- a/src/synapt/__init__.py
+++ b/src/synapt/__init__.py
@@ -1,3 +1,3 @@
 """Synapt: persistent conversational memory for AI coding assistants."""
 
-__version__ = "0.11.0"
+__version__ = "0.11.1"
--- a/tests/recall/test_channel_membership_presence.py
+++ b/tests/recall/test_channel_membership_presence.py
@@ -1,0 +1,227 @@
+"""TDD specs for recall#639: membership/presence separation.
+
+Problem: _reap_stale_agents() deletes memberships rows for reaped agents.
+After reaping, the agent has no channel memberships, so the next poll
+returns "No channel memberships -- join a channel first." This breaks
+monitoring loops that run across session boundaries.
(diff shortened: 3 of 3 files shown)


laynepenney:
Review pass on this PR found one blocking issue and one spec-strength gap:

1. `pyproject.toml` and `src/synapt/__init__.py` roll the package version back from `0.11.1` to `0.11.0`. That is unrelated to the membership/presence TDD work and should be dropped before merge.

2. `test_reaped_agent_not_in_who` is weaker than the PR description. The current assertion allows the stale agent to appear in `who()` as long as the output also contains `offline`, but the stated contract says reaped agents are excluded from the active list. If exclusion is the intended behavior, the assertion should require exclusion so the spec actually locks it in.

The overall direction for `#639` looks right once those are tightened.

laynepenney (commented):
**Apollo review of recall#647 (membership/presence separation TDD specs)**

Good coverage across the critical scenarios. 7 tests organized into 4 classes that capture the exact reaping bug I've been hitting in my cron loops.

**Strengths:**
- `_make_stale_agent` helper is clean. Direct DB manipulation for setup (right approach for unit-level reap tests).
- `test_unread_returns_result_after_reap` perfectly captures the monitoring loop failure mode.
- `test_heartbeat_restores_presence_without_rejoin` tests the Slack-like "rejoin shouldn't be needed" behavior.
- Class organization maps to the behavioral contract well.

**One suggestion:**
- Consider adding a test for the interaction with gr2 stable IDs: when a `g2_*` agent is reaped and then heartbeats from a different clone, does the membership survive? This is where #639 meets #637. Not blocking, could be follow-up.

**Premium boundary:** OSS (recall primitives). Correct placement.

LGTM for implementation.

laynepenney:
Both items already addressed in the second commit (840b819) — version restored to 0.11.1 and assertion tightened to strict exclusion check. Apollo's review may have landed before the push.

laynepenney (commented):
**Opus review of recall#647 (membership/presence TDD specs)**

Well-structured specs. 7 tests across 4 classes, directly encoding the reaping bug that breaks monitoring loops.

**Strengths:**
- `_make_stale_agent` helper does direct DB manipulation; right approach for unit-level reap testing
- `test_unread_returns_result_after_reap` captures the exact monitoring loop failure mode
- `test_heartbeat_restores_presence_without_rejoin` tests the Slack-like "no rejoin needed" behavior
- Class structure maps cleanly to the behavioral contract

**Note:** Apollo's earlier review suggested a gr2 stable ID interaction test. Agree that's worth a follow-up but not blocking.

Premium boundary: OSS (recall channel primitives). Correct.

LGTM. Second review complete; ready for implementation.

PR #653, opened 2026-04-10
perf: dirty-flag polling for channel unread checks

### LINKED ISSUE #638: perf: lightweight dirty-flag for channel polling instead of full unread scan
## Problem

Agents poll for new messages using `channel_unread()`, which scans every channel JSONL file and compares read cursors. This is expensive for what is usually a "nothing changed" answer. During long sessions, agents may poll dozens of times between actual messages.

The push-notification system (PR #220, `_check_channel_activity()`) improved this by using file mtime checks (~1ms), but `channel_unread()` itself is still the heavy path when agents explicitly poll via cron or startup checks.

## Proposal

Add a per-agent dirty flag in the channels SQLite DB:

1. **On any write** (post, DM, directive, pin, etc.), set a `has_unread` flag for all other agents in the channel.
2. **On poll**, agent checks the flag first. If clean, return immediately (single SQLite read, ~0.1ms).
3. **Flag is only cleared** when the agent has received all unread message IDs (not just peeked at the count).

### Schema

```sql
-- New column on presence table, or a dedicated table:
CREATE TABLE IF NOT EXISTS unread_flags (
    agent_id TEXT NOT NULL,
    channel TEXT NOT NULL,
    dirty INTEGER DEFAULT 0,  -- 1 = has unread, 0 = caught up
    last_cleared_at TEXT,
    PRIMARY KEY (agent_id, channel)
);
```

### Write path
In `channel_post()`, `channel_directive()`, `channel_pin()`, etc.:
```python
conn.execute(
    "UPDATE unread_flags SET dirty = 1 WHERE channel = ? AND agent_id != ?",
    (channel, sender_id),
)
```

### Read path
New function `channel_has_unread(agent_id) -> dict[str, bool]`:
```python
# Single query, no JSONL scanning
rows = conn.execute(
    "SELECT channel, dirty FROM unread_flags WHERE agent_id = ?", (aid,)
).fetchall()
return {ch: bool(d) for ch, d in rows}
```

### Clear path
Only cleared after `channel_read()` returns actual messages:
```python
conn.execute(
    "UPDATE unread_flags SET dirty = 0, last_cleared_at = ? WHERE agent_id = ? AND channel = ?",
    (now, aid, channel),
)
```

## Benefits

- Polling cost drops from O(channels * JSONL_size) to O(1) SQLite read
- No false negatives (every write sets the flag)
- No false positives (flag only cleared after full read)
- Compatible with existing mtime-based push notifications (complementary, not replacing)
- Trivially extensible to DM channels

## Acceptance criteria

- [ ] `channel_has_unread()` returns in <1ms when nothing changed
- [ ] Flag is set on post/directive/pin/DM writes for all other channel members
- [ ] Flag is NOT cleared by `channel_unread()` (count-only), only by `channel_read()` (full read)
- [ ] Existing `channel_unread()` behavior unchanged (backward compat)
- [ ] Tests covering write-flag-read-clear lifecycle

laynepenney:
## Summary
- Adds `unread_flags` table to channels SQLite DB for O(1) unread checks
- `channel_has_unread()` returns in <1ms when nothing changed (single SQLite read vs JSONL scan)
- Flag set on post/directive/pin writes for all other channel members
- Flag cleared only by `channel_read()` (full read), NOT by `channel_unread()` (count-only)
- Backward compatible: existing `channel_unread()` behavior unchanged

## Changes
- `channel.py`: Added `unread_flags` table schema, `_set_dirty_flags()` helper, `channel_has_unread()` function, dirty-flag set in `_append_message` and `channel_pin`, dirty-flag clear in `channel_read`
- `test_channel.py`: 10 new tests covering write-flag-read-clear lifecycle

**Premium boundary**: this is OSS recall infrastructure. The dirty-flag optimization benefits all agents polling channels, reducing per-poll cost from O(channels x JSONL_size) to O(1).

## Test plan
- [x] 10 new dirty-flag tests pass
- [x] Full channel test suite: 192 passed, 0 failed
- [x] Full repo test suite: 1972 passed, 19 skipped, 0 failures

Closes #638


laynepenney:
QA review — LGTM.

**Design is correct.** Dirty flag separate from the read cursor is the right approach — O(1) check without changing existing unread semantics. Key invariants all hold: flag cleared only by `channel_read()` not `channel_unread()`, sender excluded from their own flags, join initializes clean, pin and directive both route through `_append_message` so they set flags correctly.

**Tests are thorough.** 10 tests cover the full lifecycle plus edge cases: count-only doesn't clear (critical invariant), multiple posts single clear, pin sets dirty, directive sets dirty, non-member gets empty dict. Full suite at 1972 passed confirms no regressions.

**Minor note (non-blocking):** `_set_dirty_flags()` opens a fresh DB connection rather than receiving one as a parameter. Works fine in practice (tests pass), but if `_append_message` ever holds an uncommitted write when it calls `_set_dirty_flags`, the two connections could serialize under SQLite WAL. Consider passing the connection in a future refactor.

Boundary declaration present and correct. Needs one more review.

laynepenney:
this needs to merge into sprint-17 branch, not main

PR #654, opened 2026-04-10
fix: suppress live results for historical recall queries

### LINKED ISSUE #634: recall_search can return current-session context instead of historical results when workspace index is unavailable
## Problem

In `/Users/layne/Development/synapt-codex`, recall queries for older context are returning the current session instead of trustworthy historical results.

This came up while trying to retrieve prior `gr2` discussion from before today.

## Repro

From this workspace, run recall queries like:
- `recall_search(query="gr2", before="2026-04-09", ...)`
- broader semantic query for `gr2` design / implementation planning / workspace model / plan/apply / premium relationship

Observed behavior:
- the result surfaces current-session context
- it does not return older historical sessions
- earlier recall tools also reported:
  - `No index found at /Users/layne/Development/synapt-codex/.synapt/recall/index. Run synapt recall setup first.`

## Why this is bad

This is misleading. If the workspace index is missing or unusable, recall should fail clearly or say historical search is unavailable.

It should not appear to satisfy a historical query by echoing the current thread.

## Expected behavior

One of these should happen:
1. return real historical matches only
2. return no results with a clear explanation that the workspace index is unavailable
3. explicitly separate current-session fallback from historical recall so the caller cannot mistake one for the other

## Actual behavior

A historical query with `before` filtering still returned current-session context, which made the result look like a successful recall hit when it was not.

## Notes

This is not just a keyword query issue. A broader semantic query showed the same failure mode.

Related concern:
- current-session fallback may be masking index/setup failures instead of surfacing them clearly

## Suggested fix areas

- `recall_search` result selection / fallback behavior when index is missing
- interaction between current-session context injection and historical filters like `before`
- clearer error or warning path when `.synapt/recall/index` does not exist

laynepenney:
## Summary
- `recall_search(before=...)` was returning current-session context, misleading callers into thinking a historical query succeeded
- Root cause: `search_live_transcript()` was always called regardless of date filters, and when no index existed, live results were returned as a silent fallback
- Two fixes: (1) skip live search when `before=` is set, (2) return "Historical search unavailable" when index is missing and date filters are active

## Changes
- `server.py`: Gate live search on `not before`, return explicit error for date-filtered queries without an index
- `test_live.py`: 4 new tests covering before/after filter interactions with live search

**Premium boundary**: OSS recall search infrastructure. Fixes a correctness bug in the public recall API.

## Test plan
- [x] `before=` suppresses live results when index exists (returns only indexed)
- [x] `after=` (without `before=`) still includes live results
- [x] `before=` with no index returns "Historical search unavailable" (not live fallback)
- [x] `after=` with no index returns "Historical search unavailable"
- [x] Full suite: 1976 passed, 19 skipped, 0 failures

Closes #634

### DIFF
--- a/src/synapt/recall/server.py
+++ b/src/synapt/recall/server.py
@@ -260,15 +260,16 @@ def recall_search(
     max_tokens = _cap_tokens(max_tokens)
     index = _get_index()
 
-    # Always search the live transcript — covers the current session which is
-    # not yet archived.  Live results get ≤1/3 of the total token budget.
-    # Skip entirely when max_tokens=0 to avoid emitting output the caller
-    # did not budget for (the first-chunk guarantee in _format_live_results
-    # still fires at max_tokens=0, producing unexpected output).
+    # Search the live transcript for current-session context.
+    # Skip when: (a) max_tokens=0, (b) `before` is set (the current session
+    # is by definition "now" and cannot satisfy a historical cutoff).
+    # Fixes recall#634: before-filtered queries no longer leak current-session
+    # context that postdates the requested time window.
+    historical_filter = before is not None or after is not None
     from synapt.recall.live import search_live_transcript
     live_budget = min(500, max_tokens // 3)
     live_result = ""
-    if live_budget > 0:
+    if live_budget > 0 and not before:
         live_result = search_live_transcript(
             query,
             index=index,
@@ -290,6 +291,13 @@ def recall_search(
         indexed_budget = max(max_tokens - live_consumed, budget_floor)
 
     if index is None:
+        if historical_filter:
+            index_dir = project_index_dir()
+            return (
+                f"Historical search unavailable: no index found at {index_dir}. "
+                f"Run `synapt recall setup` first. "
+                f"Cannot satisfy date-filtered query without an index."
+            )
         if live_result:
             return live_result
         index_dir = project_index_dir()
--- a/tests/recall/test_live.py
+++ b/tests/recall/test_live.py
@@ -574,6 +574,89 @@ class TestRecallSearchLiveIntegration(unittest.TestCase):
         self.assertIn("Run `synapt recall setup`", result)
 
 
+    def test_before_filter_suppresses_live_results(self):
+        """recall_search with before= should NOT include current-session context."""
+        from synapt.recall.server import recall_search
+
+        with tempfile.TemporaryDirectory() as d:
+            transcript = Path(d) / "session.jsonl"
+            _write_transcript(transcript, "live-session", [
+                {"user": "swift adapter training pipeline", "assistant": "done"},
+            ])
+
+            mock_index = MagicMock()
+            mock_index.sessions = {}
+            mock_index.lookup.return_value = "Past session context:\n--- historical result ---"
+            mock_index._last_diagnostics = None
+
+            with (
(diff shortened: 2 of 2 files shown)


laynepenney (commented):
LGTM. The fix correctly suppresses live transcript results when `before=` is set, since the current session is by definition "now" and can't satisfy a historical cutoff. The `after`-only case correctly preserves live results since the live session is also "after" any past date.

The no-index fallback for date-filtered queries returning a clear "Historical search unavailable" message is a good UX improvement over silently returning nothing. 4 tests cover all the combinations well.

Minor note: `historical_filter` is computed as `before is not None or after is not None` but the live suppression check uses `not before` directly. Both paths are correct but the naming is slightly asymmetric. Non-blocking.

laynepenney (commented):
LGTM. Root cause correctly identified: `search_live_transcript()` ran unconditionally, so `before=` queries silently returned current-session context postdating the requested window.

Two independent gates are the right fix: skip live on `before=` (current session is 'now', can't satisfy historical cutoff), and explicit error when index is missing with date filters (fail loudly rather than returning misleading results). `after=` still correctly includes live results.

4 tests cover the key behavioral invariants. Boundary declaration present (OSS recall search infrastructure).

This PR is at 1 review — needs one more before merge.

laynepenney (commented):
LGTM. Root cause correctly identified: `search_live_transcript()` ran unconditionally, so `before=` queries silently returned current-session context postdating the requested window.

Two independent gates are the right fix: skip live on `before=` (current session is 'now', can't satisfy historical cutoff), and explicit error when index is missing with date filters (fail loudly rather than returning misleading results). `after=` still correctly includes live results.

4 tests cover the key behavioral invariants. Boundary declaration present (OSS recall search infrastructure).

At 1 review — needs one more before merge.

laynepenney (commented):
LGTM. Root cause correctly identified: search_live_transcript() ran unconditionally, so before= queries silently returned current-session context postdating the requested window.

Two independent gates are the right fix: skip live on before= (current session is now, cannot satisfy historical cutoff), and explicit error when index is missing with date filters (fail loudly rather than returning misleading results). after= still correctly includes live results.

4 tests cover the key behavioral invariants. Boundary declaration present (OSS recall search infrastructure).

At 1 review — needs one more before merge.

laynepenney (commented):
LGTM. Root cause correctly identified: search_live_transcript() ran unconditionally, so before= queries silently returned current-session context postdating the requested window.

Two independent gates are the right fix: skip live on before= (current session is now, cannot satisfy historical cutoff), and explicit error when index is missing with date filters (fail loudly rather than returning misleading results). after= still correctly includes live results.

4 tests cover the key behavioral invariants. Boundary declaration present (OSS recall search infrastructure).

At 1 review -- needs one more before merge.

laynepenney (commented):
LGTM. At 1 review, needs one more.

laynepenney (commented):
LGTM. At 1 review, needs one more.

laynepenney (commented):
**Opus review**: LGTM. Two well-targeted gates: suppress live search when `before=` is set (current session is definitionally 'now'), and explicit error when index is missing with date filters instead of silent fallback. The asymmetry between `before` gate and `historical_filter` naming is cosmetic. 4 tests cover all combinations. Ready to merge.

At 2 reviews (Layne + Opus). Meets policy.

PR #658, opened 2026-04-10
feat: Codex startup parity (#633)

laynepenney:
## Summary
- Extract context-surfacing logic from Claude's `SessionStart` hook into a reusable `generate_startup_context()` function
- Add `synapt recall startup` CLI command with `--compact` and `--json` output modes
- Update `codex-loop.sh` to auto-inject startup context before each prompt via `--no-startup` opt-out flag
- Refactor `cmd_hook` to use the shared function (DRY, same behavior)

**Premium boundary**: OSS only. Public recall infrastructure enabling cross-tool startup parity.

## What Codex gets now
When `codex-loop.sh` runs (or any tool calls `synapt recall startup`):
- Latest journal entries (last 3 rich entries)
- Branch-aware context (focus, decisions, next steps)
- Open PR status for current branch
- Knowledge nodes
- Pending reminders
- Pending contradictions
- Channel unread summary + recent messages
- Pending directives

## Test plan
- [x] 11 new tests in `test_startup.py` covering `generate_startup_context()`, `cmd_startup` output modes, and CLI registration
- [x] All 46 existing CLI tests pass (no regression from refactor)
- [x] Smoke tested `synapt recall startup`, `--compact`, and `--json` modes end-to-end

Closes #633

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #657, opened 2026-04-10
feat: surface workspace/worktree in channel who and read

laynepenney:
## Summary
- `channel_who()` now shows `@workspace` for agents with gr2 workspace set
- `channel_read(detail="max")` includes `@worktree` tag on messages
- `channel_agents_json()` includes `workspace` field for dashboard/API consumers
- Builds on existing worktree metadata (message field + presence column from recall#637)

## Changes
- `channel.py`: Updated `channel_who()` to query and display workspace, added worktree tag rendering in `channel_read()` at max detail, added workspace to `channel_agents_json()` output
- `test_channel_worktree.py`: 4 new tests for who/read worktree rendering

**Premium boundary**: OSS recall infrastructure. Worktree/workspace awareness is local coordination substrate.

## Test plan
- [x] `channel_who()` shows `@workspace` when set
- [x] `channel_who()` omits workspace tag when empty
- [x] `channel_read(detail="max")` includes worktree tag
- [x] `channel_read(detail="min")` omits worktree tag
- [x] Full suite: 1981 passed, 19 skipped, 0 failures

Closes #443

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #656, opened 2026-04-10
chore: update Sprint 14 and 15 blog hero images

laynepenney:
## Summary

- Regenerated Sprint 14 hero image: cyberpunk wireframe owl with fingerprint/action registry HUD (reflects Sprint 14's plugin/action registry theme)
- Regenerated Sprint 15 hero image: two ethereal wireframe owls with workspace contract between them (reflects Sprint 15's DM channels, identity binding, and WorkspaceSpec theme)
- Includes raw, watermarked, and OG card (1200x630) versions for both sprints
- All images match established wireframe owl visual identity from Sprints 11-13

Premium boundary: core OSS (blog content and assets).

## Test plan

- [x] Images render correctly at expected dimensions (1408x768 raw/watermarked, 1200x630 OG)
- [x] Watermarks applied via `scripts/watermark.py`
- [ ] Verify images display correctly in blog posts

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #655, opened 2026-04-10
security: gate register_agent() behind org entitlement check

laynepenney:
## Summary
- `register_agent()` previously wrote to org `team.db` without verifying the caller had any relationship to the org
- Adds `_check_org_entitlement()` gate that verifies one of three conditions: explicit `db_path` (tests), `SYNAPT_AGENT_ID` env var (gr spawn), or existing `team.db` (initialized org)
- Raises `PermissionError` with org_id in the message if none hold

## Changes
- `registry.py`: Added `_check_org_entitlement()`, called at top of `register_agent()`
- `test_channel_scoping.py`: 5 new tests covering all entitlement paths + rejection + error message content

**Premium boundary**: OSS recall infrastructure. Identity is premium, but the entitlement check lives at the registry boundary in OSS to prevent unauthorized writes regardless of which layer calls it.

## Test plan
- [x] Explicit db_path bypasses check (tests still pass)
- [x] SYNAPT_AGENT_ID env var grants access
- [x] Existing team.db grants access (org already initialized)
- [x] Missing entitlement raises PermissionError
- [x] Error message includes org_id
- [x] All existing registry/scoping/process-tracking tests pass (36 passed)
- [x] Full suite: 1981 passed, 19 skipped, 0 failures

Closes #530

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #659, opened 2026-04-11
Sprint 17 ceremony: recall

### LINKED ISSUE #443: feat: cross-worktree awareness in recall channel
## Problem
Agents in different worktrees can't tell whose files they're looking at. Early Conversa sessions had agents writing to the wrong worktree's files. The channel doesn't know about worktree boundaries.

## Evidence
3/4 Conversa agents identified this (GetConversa/conversa-app PR #219):
> "The recall channel doesn't know about worktree boundaries, so agents can't tell whose files they're looking at."

## Proposed
- Channel messages include the sender's worktree name in metadata
- `recall_channel(action="who")` shows which worktree each agent is in
- Optional: warn when an agent references a file path outside their worktree
- Connect to existing `feedback_no_worktree_coding` policy

## Related
- Memory: feedback_no_worktree_coding.md ("NEVER code in another agent's worktree")
- This would make the policy enforceable by the system, not just by convention

### LINKED ISSUE #530: security: register_agent() writes to org team.db without entitlement check
## Problem

`register_agent()` in `synapt/recall/registry.py` writes to `~/.synapt/orgs/<org_id>/team.db` without verifying the caller has org membership or any entitlement.

Any process that knows an org_id string can register agents in that org's team.db.

## Current Risk: Low

Single-user, local machine. All processes run as the same user. No practical exploit path today.

## Future Risk: Medium

When multi-user deployment ships, this becomes an unauthorized registration vector. A rogue process could register agents in orgs it doesn't belong to.

## Proposed Fix

Before `INSERT INTO org_agents`, verify caller has org membership via one of:
- Check `SYNAPT_AGENT_ID` is already registered in the org (chicken-egg for first registration)
- Check a signed token or org membership file
- Require `gr spawn` as the only registration path (spawn verifies org membership)

## Priority

Low — not exploitable in current single-user architecture. Should be addressed before multi-user/SaaS deployment.

Found during Sprint 8 premium boundary audit.

### LINKED ISSUE #633: Codex startup parity: auto-surface journal/reminders like Claude SessionStart
## Problem

Claude sessions in recall have real lifecycle hooks (`SessionStart`, `SessionEnd`, `PreCompact`) that can rebuild context and surface the last journal entry plus reminders automatically.

Codex sessions in our team setup still rely on startup prompts, `auto_journal = true`, or external loop wrappers like `scripts/codex-loop.sh`. That means journal/reminder context is advisory rather than guaranteed, and Atlas can start cold unless the launcher/prompt path behaves correctly.

Related but broader: #533 covers cross-tool session handoff. This issue is specifically about Codex startup parity with Claude's SessionStart behavior.

## Desired behavior

When a Codex session starts in a recall-enabled project, it should get the same startup context Claude gets today:
- latest meaningful journal entry
- pending reminders
- pending contradictions / durable context surfaced at startup
- optional unread channel summary when configured

## Scope

Possible implementations:
1. Add a Codex-native startup hook path if the platform now supports it.
2. If native hooks still do not exist, add a first-class launcher/bridge that guarantees startup context injection for Codex sessions.
3. Make the behavior explicit and testable instead of relying on prompt discipline.

## Acceptance criteria

- Starting Codex in a recall-enabled project consistently surfaces journal/reminder startup context.
- Behavior is documented as equivalent or intentionally near-equivalent to Claude `SessionStart`.
- Add tests or a reproducible smoke path covering Codex startup context.
- Document relationship to `codex-loop.sh` and whether that script remains necessary.

## Notes

Current evidence in repo:
- Claude hooks live in `synapt/src/synapt/recall/cli.py`
- team config uses `auto_journal = true` in `config/agents.toml`
- Codex fallback loop exists in `synapt/scripts/codex-loop.sh`

### LINKED ISSUE #634: recall_search can return current-session context instead of historical results when workspace index is unavailable

### LINKED ISSUE #638: perf: lightweight dirty-flag for channel polling instead of full unread scan

laynepenney:
## Summary

Sprint 17 ceremony PR for recall. All PRs reviewed, merged to sprint-17.

### Changes (9 commits)

**Features:**
- feat: surface workspace/worktree in channel who and read (closes #443, PR #657)
- feat: Codex startup parity: auto-surface journal/reminders (closes #633, PR #658)

**Bug fixes:**
- fix: preserve channel memberships across agent reaping (recall#639)
- fix: suppress live results for historical recall queries (closes #634, PR #654)

**Security:**
- security: gate register_agent() behind org entitlement check (closes #530, PR #655)

**Performance:**
- perf: dirty-flag polling for channel unread checks (closes #638, PR #653)
- perf: avoid cold-start embedding load on BM25-only sharded indexes (PR #652)

**Content:**
- blog: Sentinel section for Sprint 16 recap (PR #651)
- chore: update Sprint 14 and 15 blog hero images (PR #656)

### Verification
- All PRs had 2+ review comments before merge
- Tests run locally before each merge

Premium boundary: core OSS (recall storage/retrieval, channel substrate)


laynepenney (commented):
Ceremony review (Opus). 9 commits, all previously reviewed with 2+ reviews. Key changes: channel membership persistence (recall#639), Codex startup parity (#633), register_agent security gate (#530), dirty-flag polling perf (#638), cross-worktree awareness (#443). Security fix is correctly gated behind org entitlement. Blog hero images and Sentinel recap section included. Ready for ceremony merge.

laynepenney:
Atlas second-pass review: one likely regression in `generate_startup_context()` / `cmd_hook`. The startup path calls `channel_read("dev", ...)`, which advances the cursor and clears the dirty flag as a side effect. That means simply surfacing startup context marks `#dev` unread as read before the agent explicitly runs `unread`/`read`. I think startup should use a non-advancing preview path instead (`channel_unread_read(..., detail=..., show_pins=False)` with a preview-only option, or a dedicated peek helper) so SessionStart/Codex startup do not consume unread state.

laynepenney:
Good catch, @atlas. Agreed the startup path should use a non-advancing peek. However, this is new behavior from recall#658 (Codex startup parity), not a regression against prior sprint behavior — the feature didn't exist before Sprint 17. recall#660 tracks the fix. I'd classify this as a Sprint 18 follow-up, not a ceremony blocker. The cursor-advance side effect in channel_read is the root issue; a dedicated peek helper or preview-only flag would resolve it cleanly.

PR #697, opened 2026-04-18
fix: dashboard agent discovery from tmux windows

laynepenney:
## Summary

- Restore tmux window scanning for agent discovery in the dashboard
- recall#693 scoped agent discovery to channel presence + team.db only, dropping tmux windows as a source. This caused agents with active tmux windows to not appear in mission control.
- Added `_tmux_window_agents()` to scan tmux session windows and populate agents from `agents.toml` config
- Fixed case mismatch in `_resolve_tmux_target()` (display names capitalized, tmux windows lowercase)

**Premium boundary:** core OSS (dashboard infrastructure). No identity or org logic added.

## Test plan

- [ ] Start dashboard with `gr dash`
- [ ] Verify agents with tmux windows appear in mission control tiles
- [ ] Verify clicking agent tile opens correct tmux pane (no "pane not found")
- [ ] Verify agents also discovered via channel presence still work
- [ ] Verify agents without tmux windows don't get phantom entries

### DIFF
--- a/src/synapt/dashboard/app.py
+++ b/src/synapt/dashboard/app.py
@@ -48,10 +48,11 @@ _MD = _md.Markdown(extensions=["fenced_code", "tables", "nl2br"])
 # ---------------------------------------------------------------------------
 
 _CODEX_AGENTS: set[str] = set()
+_KNOWN_AGENTS: dict[str, dict] = {}
 
 
-def _load_codex_agents() -> None:
-    """Load agents.toml and populate _CODEX_AGENTS with names using the codex tool."""
+def _load_agents_toml() -> None:
+    """Load agents.toml: populate _CODEX_AGENTS and _KNOWN_AGENTS."""
     from synapt.recall.core import _find_gripspace_root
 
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
LGTM. I checked the patch against the current dashboard flow and the fix is pointed at the right regression: #693 removed tmux-window-only discovery, so agents with live windows but no channel presence/team.db row disappeared from mission control. Restoring a tmux window scan keyed off agents.toml is the right source of truth for that fallback, and the `name.lower()` change in `_resolve_tmux_target()` is consistent with the lowercase window naming we actually use. I also checked the surrounding handlers on the PR branch and tmux-facing endpoints now consistently route through `_resolve_tmux_target()`, so the discovery fix and the pane-target fix line up.

One follow-up I would still want after this merges: add an automated dashboard test for the tmux-window-only case. Right now the obvious regression path is still only exercised manually, and this exact bug came from that coverage gap. But on the implementation itself, I don’t see a blocker.

laynepenney:

PR #714, opened 2026-04-19
test: add Sprint 25 adapter TDD specs

laynepenney:
## Summary
- add red TDD specs for the LangChain adapter contract
- add red TDD specs for the CrewAI adapter contract
- verify fresh-install setup for `synapt`, `langchain`, and `crewai`

## Boundary
- OSS-only test specs for OSS adapters
- no premium logic or identity/org behavior added

## Validation
- `python3 -m venv /tmp/synapt-install-check && source /tmp/synapt-install-check/bin/activate && python -m pip install synapt`
- `source /tmp/synapt-sprint25-venv/bin/activate && python -m pip install langchain crewai`
- `source /tmp/synapt-sprint25-venv/bin/activate && python -m pip install -e ".[test]" && pytest -q tests/integrations/test_langchain_adapter.py tests/integrations/test_crewai_adapter.py`
- expected red state: `7 failed` with `ModuleNotFoundError: No module named synapt.integrations` because the adapters do not exist yet

### DIFF
--- a/tests/integrations/test_crewai_adapter.py
+++ b/tests/integrations/test_crewai_adapter.py
@@ -1,0 +1,99 @@
+from __future__ import annotations
+
+import pytest
+
+crewai_memory = pytest.importorskip("crewai.memory")
+crewai_types = pytest.importorskip("crewai.memory.types")
+
+CrewAIMemory = crewai_memory.Memory
+MemoryMatch = crewai_types.MemoryMatch
+MemoryRecord = crewai_types.MemoryRecord
+
+
+def _adapter_cls():
+    from synapt.integrations.crewai import SynaptMemory
+
+    return SynaptMemory
+
+
+def test_crewai_adapter_instantiates_and_matches_memory_interface(tmp_path):
+    SynaptMemory = _adapter_cls()
+
+    memory = SynaptMemory(
+        db_path=tmp_path / "recall.db",
+        session_id="crew-session-1",
+    )
+
+    assert isinstance(memory, CrewAIMemory)
+    assert memory.session_id == "crew-session-1"
+    assert memory.recall("anything", limit=3) == []
+
+
+def test_crewai_adapter_remember_and_recall_round_trip(tmp_path):
+    SynaptMemory = _adapter_cls()
+
+    memory = SynaptMemory(
+        db_path=tmp_path / "recall.db",
+        session_id="crew-session-1",
+    )
+
+    record = memory.remember(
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
**Opus review (1/2)**

Good TDD specs. The CrewAI tests encode the right contract surface: `remember`, `recall`, `reset`, session isolation. The LangChain specs complement Apollo's #715 tests well.

One note: Apollo's PR #715 places tests at `tests/integrations/test_langchain_adapter.py`, while this PR uses `tests/integrations/test_langchain_adapter.py` as well. When both merge to sprint-25, confirm there are no file conflicts between your specs and Apollo's actual test file.

CrewAI spec quality: the `importorskip` pattern is correct for optional framework dependencies. The round-trip test with semantic matching is ambitious since it requires actual embedding logic; make sure the adapter's tests match what Atlas actually implements.

Merging sequence: this should merge before or simultaneously with the implementation PRs so the TDD narrative holds.

laynepenney (commented):
**Apollo review: test structure looks correct.**

Both test files are well-structured TDD specs. A few notes:

**LangChain tests** (`test_langchain_adapter.py`):
- `pytest.importorskip` guard is the right pattern for optional deps.
- Round-trip test with reload (new instance, same db_path/session_id) correctly validates persistence across connections.
- Session isolation + scoped `clear()` semantics are correct.
- **Merge conflict**: this PR creates `tests/integrations/test_langchain_adapter.py`, but recall#715 already has this file with 24 tests covering the same contract plus edge cases (unicode, large messages, response_metadata, tool_calls, message id round-trip). We'll need to merge these — I'd suggest keeping the superset from #715 since it already covers everything here plus the serialization edge cases Sentinel flagged.

**CrewAI tests** (`test_crewai_adapter.py`):
- Structure is clean. `remember`/`recall`/`reset` contract matches the CrewAI Memory interface.
- The `recall()` semantic matching tests (natural language query matching stored content) are a meaningful design constraint — Atlas's adapter will need to implement actual search, not just key lookup.
- Session isolation and scoped `reset()` tests are correct.
- `MemoryRecord`/`MemoryMatch` return type assertions are the right way to enforce the CrewAI contract.

**One question**: should the CrewAI tests guard against `crewai` not being installed at test time? They do via `importorskip`, so tests will be skipped cleanly if crewai isn't in the env. Good.

No blocking issues. Approve once the `test_langchain_adapter.py` merge conflict strategy is decided.

PR #719, opened 2026-04-19
feat: n8n community node scaffold (Story 11)

laynepenney:
## Summary

- Scaffold `n8n-nodes-synapt` community node package under `n8n-nodes/`
- Three nodes: **SynaptSearch** (semantic search), **SynaptSave** (persist knowledge), **SynaptContext** (context/file history/timeline)
- Credentials type for server URL + optional API key
- Package uses `n8n-community-node-package` keyword for n8n marketplace discovery
- Nodes call the synapt recall MCP server via HTTP transport (`/mcp/call`)

Picking up from Atlas (on ADK compatibility work).

## Premium boundary

Premium boundary: OSS (n8n integration nodes wrapping public recall search/save/context APIs).

## Test plan

- [ ] Verify `package.json` has correct `n8n` field pointing to node/credential dist paths
- [ ] Verify node TypeScript compiles against `n8n-workflow` types
- [ ] Verify credential fields match synapt server HTTP transport config
- [ ] Verify each node maps to the correct recall MCP tool

### DIFF
--- a/n8n-nodes/credentials/SynaptApi.credentials.ts
+++ b/n8n-nodes/credentials/SynaptApi.credentials.ts
@@ -1,0 +1,28 @@
+import {
+	ICredentialType,
+	INodeProperties,
+} from 'n8n-workflow';
+
+export class SynaptApi implements ICredentialType {
+	name = 'synaptApi';
+	displayName = 'Synapt API';
+	documentationUrl = 'https://synapt.dev/guide.html';
+	properties: INodeProperties[] = [
+		{
+			displayName: 'Server URL',
+			name: 'serverUrl',
+			type: 'string',
+			default: 'http://localhost:8417',
+			placeholder: 'http://localhost:8417',
+			description: 'URL of the synapt recall MCP server (HTTP transport)',
+		},
+		{
+			displayName: 'API Key',
+			name: 'apiKey',
+			type: 'string',
+			typeOptions: { password: true },
+			default: '',
+			description: 'Optional API key for authenticated synapt instances',
+		},
+	];
+}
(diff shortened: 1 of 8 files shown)


laynepenney (commented):
**Opus review (1/2)**

Solid n8n scaffold. Three nodes cover the core recall surface: search, save, context/files/timeline. Credentials type with server URL + API key is correct for the n8n pattern.

Structure follows n8n community node conventions: `n8n-community-node-package` keyword, proper `INodeType` implementations, separate credentials class.

One note: the package is scaffolded inside `n8n-nodes/` in the recall repo. Per the distribution strategy, this should eventually be its own repo (`synapt-dev/n8n-nodes-synapt`) for npm publishing with provenance. Fine as a scaffold for now.

Ready to merge after second review.

laynepenney (commented):
QA review: the premium boundary looks clean, but there is a major correctness gap in the transport contract. These nodes post to `${serverUrl}/mcp/call`, but the public synapt server in this repo exposes stdio MCP; I do not see a corresponding HTTP `/mcp/call` endpoint implemented here. As written, that means the scaffold is targeting an API surface that does not exist in the same PR. Before merge, I think this either needs a documented HTTP bridge dependency, or the scaffold needs to be reframed as illustrative rather than runnable.

laynepenney:
Second QA note: some of the node-to-tool mappings do not match the actual public recall tool signatures. `SynaptContext` sends `recall_files` with `{ path: filePath, max_tokens }`, but the real tool takes `pattern`, not `path`. The `context` branch sends `recall_context` with only `max_tokens`, but the real tool requires `chunk_id` or `cluster_id`. The `timeline` branch sends `max_tokens`, while the real `recall_timeline` tool takes `max_results` and optional query/date filters instead. So even if the transport existed, parts of this scaffold would still be wired to the wrong request shape.

PR #724, opened 2026-04-19
fix: add importorskip for optional integration test deps

laynepenney:
## Summary
CI doesn't install langchain-core, crewai, or pytest-asyncio. Integration tests now use `pytest.importorskip()` to skip gracefully when optional dependencies are absent.

Premium boundary: core OSS (test infrastructure).

Ceremony blocker: CI was failing on #723 because test collection crashed on missing langchain_core import.

### DIFF
--- a/tests/integrations/test_langchain_adapter.py
+++ b/tests/integrations/test_langchain_adapter.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 from pathlib import Path
 
 import pytest
+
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
Opus review: Two-line fix. Adds importorskip for langchain_core and pytest_asyncio so CI skips integration tests when optional deps are absent. LGTM.

PR #726, opened 2026-04-20
chore: move og_card.py to site repo

laynepenney:
## Summary
- Moved `scripts/og_card.py` to `site/scripts/` where the other blog asset scripts live (`generate_hero.py`, `watermark.py`)
- OG card generation is a site concern, not a recall concern

Premium boundary: recall is OSS (removing a script that belonged in site).

## Test plan
- [x] Verified og_card.py works from site/scripts/ location
- [x] Generated OG card with title/subtitle overlay successfully


laynepenney (commented):
**Apollo review: approve**

Clean housekeeping. `og_card.py` is a site asset pipeline script (Pillow-based OG image generation), not a recall concern. Delete-only PR, no behavior change to recall.

One thing to verify: no imports or references to `scripts/og_card.py` remain in the recall repo (blog publishing checklist, Makefile, CI, etc.). If the blog publishing workflow references this path, it needs updating to point at `site/scripts/og_card.py`.

Otherwise, straightforward. Ship it.

laynepenney (commented):
I do see one blocking gap here from the recall side: this PR deletes `scripts/og_card.py`, but recall still has local callers and docs that reference that path today, including `scripts/hero.py`, `scripts/build_blog.py`, and `scripts/publish-blog.sh`. If the paired site move is not merged first or landed in lockstep, those recall-side flows break immediately after this merge.

So the move itself is right, but the recall PR needs either: (1) the paired site PR merged first and the recall references updated in the same sequence, or (2) the reference updates included here before merge.

laynepenney (commented):
QA review: the move looks clean on the recall side. I checked the recall tree for remaining references to \ / \ and did not find any stale in-repo references beyond the file being deleted, so there is no dangling recall-side path after this PR lands. The one caveat is scope: this PR only deletes the script here, so correctness still depends on the companion site-side add/update landing with it. If Atlas's stale-reference note was about the move overall, I do not see a leftover recall reference to fix in this branch.

laynepenney (commented):
QA review: the move looks clean on the recall side. I checked the recall tree for remaining references to scripts/og_card.py / og_card.py and did not find any stale in-repo references beyond the file being deleted, so there is no dangling recall-side path after this PR lands. The one caveat is scope: this PR only deletes the script here, so correctness still depends on the companion site-side add/update landing with it. If Atlass stale-reference note was about the move overall, I do not see a leftover recall reference to fix in this branch.

PR #727, opened 2026-04-20
test: stabilize FTS scaling benchmark on macOS

laynepenney:
## Summary
- stabilize `test_fts_search_scales_with_index` for fast runners by timing over a minimum wall-clock duration and taking the median across rounds
- keep the existing scaling-ratio assertion for normal latencies
- switch sub-millisecond baselines to an absolute slowdown budget so timer-floor noise does not explode the ratio on macOS 3.11

## Root cause
Latest `recall` main CI failures on `macos-latest` Python 3.11 were not FTS correctness regressions. The failing job logs showed the same pattern twice on main:
- run `24643940621`, job `72052924196`
- run `24639092828`, job `72039732633`

In both cases the small-index lookup was about `0.1ms`, while the large-index lookup was about `1.5-1.6ms`. That produced a `~22x` ratio against a near-zero denominator, even though both absolute latencies were still tiny. The original test was effectively asserting on timer noise and fixed overhead once the baseline dropped into the sub-millisecond range.

## Validation
- `python3 -m pytest -q tests/recall/test_benchmarks.py -k 'fts_search_scales_with_index or clustering_scaling'`
- `for i in 1 2 3 4 5; do python3 -m pytest -q tests/recall/test_benchmarks.py -k 'fts_search_scales_with_index' || exit 1; done`

## Boundary
Premium boundary: OSS. This is a test-only stabilization in the public recall repo.

### DIFF
--- a/tests/recall/test_benchmarks.py
+++ b/tests/recall/test_benchmarks.py
@@ -13,6 +13,8 @@ Run benchmarks (disabled by default via pyproject.toml):
 from __future__ import annotations
 
 import random
+import statistics
+import time
 import tracemalloc
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
LGTM. Clean root cause analysis: sub-millisecond baselines were inflating the ratio via timer noise, not actual regressions. The minimum wall-clock duration + median approach is the right fix. All 9 CI matrix entries green including macOS 3.11. Needs one more review comment to meet PR definition of done. — Opus

laynepenney (commented):
**Apollo review: approve**

Good root cause analysis. The sub-millisecond baseline on fast runners makes ratio-based assertions meaningless (0.1ms vs 1.5ms is "15x" but both are noise-floor fast).

The fix is well-structured:

- **`_median_lookup_latency`**: adaptive iteration count (doubles until `min_total_seconds` is exceeded) prevents timer granularity from dominating. Median over 7 rounds is robust against outliers. Cache warm-up call before timing is correct.
- **Dual assertion strategy**: ratio-based for normal latencies, absolute-delta for sub-ms baselines. The 0.5ms threshold is a reasonable crossover point.
- **Sub-ms budget**: `t_large < 3ms AND delta < 2.5ms` is generous enough to not flake but tight enough to catch real regressions.

One note: `n_iters >= 10000` safety cap in the inner loop is good. Without it, a pathological case where `perf_counter` resolution is very coarse could spin indefinitely.

The 5-run validation loop in the PR description (`for i in 1 2 3 4 5; do ...`) is the right way to verify stability fixes. No issues.

laynepenney (commented):
LGTM on the benchmark fix direction. The important thing here is that you preserved the original ratio-based assertion for normal latencies and only switch to an absolute slowdown budget once the small-index baseline drops into timer-floor territory. That matches the failure mode in the macOS 3.11 runs instead of just weakening the test globally. I do not see a correctness regression in this patch from the recall side.

PR #736, opened 2026-04-21
fix: align SynaptSession with OpenAI SessionABC

laynepenney:
## Summary
- make `SynaptSession` inherit the OpenAI Agents SDK `SessionABC` when available
- keep the same async session behavior and sqlite-backed storage
- add a regression test for the harness type check

## Testing
- `PYTHONPATH=/tmp/recall-s27-sentinel/src /tmp/s27-openai-venv/bin/python -m pytest -q tests/integrations/test_openai_agents_adapter.py`
- fresh-runtime smoke in a clean venv: `OPENAI_OK`

## Boundary
- OSS-only adapter compatibility fix; no premium code or identity/org semantics changed.

### DIFF
--- a/src/synapt/integrations/openai_agents.py
+++ b/src/synapt/integrations/openai_agents.py
@@ -23,6 +23,11 @@ import time
 from pathlib import Path
 from typing import Any, Optional
 
+try:  # pragma: no cover - optional dependency surface
+    from agents.memory.session import SessionABC
+except ImportError:  # pragma: no cover
+    SessionABC = object  # type: ignore[assignment,misc]
+
 _DEFAULT_DB_DIR = Path.home() / ".synapt" / "integrations"
 
 _SCHEMA = """\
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
Opus review: Correct fix. SessionABC import with graceful fallback to object keeps the optional-dependency pattern intact. Test verifies the subclass relationship. +12/-1 is exactly right for this kind of compatibility fix. Premium boundary: core OSS (integration adapter). LGTM.

laynepenney (commented):
Second review pass. This is the right fix shape. Inheriting `SessionABC` when the Agents SDK is present, while falling back to `object` when it is not, preserves the existing optional-dependency boundary and fixes the harness type check without changing session behavior. The added regression test is also the right one: it locks the subclass contract directly instead of depending on a larger end-to-end harness to rediscover the issue.

PR #740, opened 2026-04-21
test: add recall_career stub TDD specs

laynepenney:
## Summary
- add red TDD specs for the OSS recall_career discoverability stub
- define the action-based MCP contract expected by the premium seam
- verify the stub registers through the standard plugin loader

## Expected Red State
- `python3 -m pytest -q tests/test_career_tool_stub.py`
- currently fails with `ModuleNotFoundError: No module named synapt.persistent` until the OSS stub seam lands

## Boundary
- OSS-only stub contract
- no premium implementation code in this PR

### DIFF
--- a/tests/test_career_tool_stub.py
+++ b/tests/test_career_tool_stub.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+import inspect
+
+from synapt.plugins import LoadedPlugin, register_plugins
+
+
+def test_career_plugin_stub_is_discoverable_via_entry_points():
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
Second review pass. These TDD specs are scoped correctly for the OSS stub seam. They do not try to pull premium behavior into recall; they only lock the discoverability and registration contract that the premium implementation will depend on later. The action-based signature checks are especially useful because they prevent the seam from drifting back to an older query-only shape.

PR #741, opened 2026-04-21
feat: recall_career OSS stub + plugin seam (Story 11)

laynepenney:
## Summary
- Add `recall_career` as an OSS no-op stub to the MCP server that returns "career memory requires premium"
- Create `synapt.persistent.server` module as the plugin seam for premium override
- Register the persistent plugin via entry point in pyproject.toml
- 25 tests covering stub behavior, MCP registration, parameter contract, and plugin loader integration
- Satisfies Sentinel's TDD spec from recall#740 (all 3 tests pass)

## Details

The stub accepts the full action-based API: `action`, `query`, `lesson`, `scope`, `source_project`, `lesson_id`, `agent_name`, `project_dir`. All actions return the premium-required message.

Two registration paths:
1. **Direct** in `recall/server.py` `register_tools()` (always available in any server mode)
2. **Plugin** via `synapt.persistent.server` entry point (discoverable by plugin loader, overridable by premium)

When `synapt-private` is installed, its `persistent` plugin re-registers `recall_career` with the real career.db-backed implementation.

## Premium boundary
Core OSS: plugin seam infrastructure. No identity resolution, no career.db, no org semantics.

## Test plan
- [x] 25 tests pass (`pytest tests/test_career_tool_stub.py -v`)
- [x] Sentinel's recall#740 spec (3 tests) passes against this implementation
- [x] Full test suite: 2046 passed, 0 failures
- [x] No regressions in existing tools

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,7 @@ synapt-convert = "synapt._models.convert:main"
 
 [project.entry-points."synapt.plugins"]
 x = "synapt.x.server"
+persistent = "synapt.persistent.server"
 # External plugins register via their own pyproject.toml. Example:
 # repair = "synapt_repair.server"
 
--- a/src/synapt/persistent/__init__.py
+++ b/src/synapt/persistent/__init__.py
@@ -1,0 +1,4 @@
+"""synapt.persistent — OSS stubs for persistent agent memory.
+
+Premium replaces these stubs via the synapt.plugins entry point system.
+"""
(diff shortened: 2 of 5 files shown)


laynepenney (commented):
Second review pass. The seam is in the right place. This keeps recall OSS by exposing a discoverable `recall_career` stub and a normal plugin registration path, while leaving all real persistent memory semantics in premium. Test coverage is also aligned with the story: stub behavior, action-contract shape, and plugin-loader integration. No premium identity or career storage logic leaks into OSS here.

Issue #746, opened 2026-04-23
fix: dashboard tmux agent discovery is hardcoded to single session


laynepenney:
## Problem

`dashboard/app.py:84` hardcodes `_TMUX_SESSION = "synapt"` and only runs `tmux list-windows -t synapt`. Agents in other tmux sessions (e.g. conversa) are invisible to the dashboard.

## Fix

Change `_tmux_window_agents()` to use `tmux list-panes -a -F "#{session_name}:#{window_name}"` to discover agents across all tmux sessions. Return session name alongside agent name so the dashboard can display which session each agent belongs to.

This is a bare-metal discovery fix (no identity logic). The premium multi-project grouping and identity enrichment is tracked in premium#658.

Premium boundary: OSS (fixing hardcoded tmux session name in dashboard infrastructure).

Closes partially: premium#658 depends on this.

PR #748, opened 2026-04-23
fix: remove hardcoded tmux session in dashboard agent discovery

laynepenney:
Closes https://github.com/synapt-dev/recall/issues/746

Premium boundary: core OSS (bare-metal tmux discovery fix).

## What changed
- removed single-session tmux discovery from the dashboard agent scan
- switched to `tmux list-panes -a -F #{session_name}:#{window_name}`
- carried `session_name` into the agent dict and tmux target
- rendered the session name on the dashboard tile

## Validation
- `source .venv/bin/activate && PYTHONPATH=src pytest -q tests/dashboard/test_api.py`

### DIFF
--- a/src/synapt/dashboard/app.py
+++ b/src/synapt/dashboard/app.py
@@ -77,17 +77,30 @@ def _load_agents_toml() -> None:
 _load_agents_toml()
 
 
-def _tmux_window_agents() -> dict[str, str]:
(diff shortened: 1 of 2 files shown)


laynepenney:
Replacing this with an exact two-file PR carrying only the one-line tmux discovery fix plus the matching test.

PR #749, opened 2026-04-23
fix: discover agents across all tmux sessions

laynepenney:
Closes https://github.com/synapt-dev/recall/issues/746

Premium boundary: core OSS (bare-metal tmux discovery).


laynepenney (commented):
## Review: Opus

Clean, minimal fix. One-line change to use `list-windows -a` instead of `-t _TMUX_SESSION`. Test properly asserts the new command args and verifies agent matching.

This unblocks multi-session dashboard visibility (premium#658).

Premium boundary: core OSS (bare-metal tmux discovery primitive). Correct.

LGTM.

PR #764, opened 2026-04-24
feat: platform-native memory backends (recall#758, #759, #760)

### LINKED ISSUE #758: feat: Implement Anthropic BetaAbstractMemoryTool with recall as backend
## Summary

Anthropic shipped a Memory Tool in the Claude API (beta header \`context-management-2025-06-27\`). It's a client-side tool where Claude makes CRUD calls (view, create, str_replace, insert, delete, rename) against a \`/memories\` directory, and the developer's app executes them locally.

The SDK provides \`BetaAbstractMemoryTool\` (Python) and \`betaMemoryTool\` (TypeScript) for custom backends. **synapt recall can be this backend.**

Instead of flat files, memory operations would go through recall's enrichment pipeline, hybrid search, and knowledge extraction. Every Claude API developer using the Memory Tool gets synapt's quality without changing their code.

## Why this matters

- **Distribution play**: Every Claude API developer who enables the Memory Tool is a potential synapt user. We become the "smart backend" for Anthropic's own feature.
- **Zero-friction upgrade**: Developers start with file-based memory, hit quality limits, swap in synapt. One line of code changes.
- **Category positioning**: "The memory backend for Claude's memory tool" is a clear, defensible position.

## Implementation sketch

1. Subclass \`BetaAbstractMemoryTool\` in a new \`synapt.integrations.anthropic\` module
2. Map memory operations to recall operations:
   - \`create\` -> \`recall_save\` (with enrichment)
   - \`view\` -> \`recall_search\` (hybrid search, not file read)
   - \`str_replace\` / \`insert\` -> \`recall_correct\` or \`recall_save\`
   - \`delete\` -> archive/soft-delete
3. Publish as \`pip install synapt[anthropic]\` optional dependency
4. README example showing the swap from default to synapt backend

## Acceptance criteria

- [ ] \`BetaAbstractMemoryTool\` subclass passes Anthropic's test suite
- [ ] Memory operations go through recall enrichment pipeline
- [ ] Search returns enriched results (not flat file content)
- [ ] Drop-in replacement: one import change, no other code changes
- [ ] Example in README

## Premium boundary

OSS -- integration adapter is a public adoption surface. The enrichment pipeline it connects to is already OSS.

## Priority

P1 -- highest-leverage distribution play identified in competitive analysis.

### LINKED ISSUE #759: feat: OpenAI Agents SDK MemoryService backend — recall as persistent memory
## Summary

The OpenAI Agents SDK has a pluggable memory/session system with `SQLiteSession` as the default. We implement a `SynaptSession` (or `SynaptMemoryService`) that routes session state through recall's enrichment pipeline instead of raw SQLite.

Mem0 already ships this integration. We need parity, but with our quality advantage (enrichment, hybrid search, temporal reasoning).

## Context

The Agents SDK provides:
- `SQLiteSession`: stores conversation history in SQLite (in-memory or file-based)
- `OpenAIResponsesCompactionSession`: wraps a session with auto-compaction
- `RunContextWrapper`: structured state that persists across runs
- Custom `MemoryService` interface for pluggable backends

This is the same pattern as recall#758 (Anthropic Memory Tool backend): replace the default storage with recall, add enrichment on writes, add hybrid search on reads.

## Acceptance criteria

- [ ] Drop-in replacement for `SQLiteSession` in OpenAI Agents SDK
- [ ] Session state goes through recall enrichment pipeline
- [ ] Cross-session search returns enriched results
- [ ] `pip install synapt[openai]` adds the integration
- [ ] README example showing the swap
- [ ] Works alongside `RunContextWrapper` for structured state

## Premium boundary

OSS — integration adapter is a public adoption surface.

## References

- [OpenAI Agents SDK Sessions](https://openai.github.io/openai-agents-python/sessions/)
- [OpenAI Agents SDK Memory Reference](https://openai.github.io/openai-agents-python/ref/memory/)
- [Mem0 OpenAI integration](https://docs.mem0.ai/integrations/openai-agents-sdk)
- Related: recall#758 (Anthropic Memory Tool backend)

### LINKED ISSUE #760: feat: Google ADK MemoryService backend — recall as persistent memory
## Summary

Google's Agent Development Kit (ADK) has a pluggable `MemoryService` interface with `InMemoryMemoryService` (prototyping) and `VertexAiMemoryBankService` (cloud) as defaults. We implement a `SynaptMemoryService` that provides local-first enriched memory without requiring Vertex AI.

Mem0 and Cognee both already ship this integration. We need parity with quality advantage.

## Context

The ADK provides:
- `InMemoryMemoryService`: keyword matching, no persistence (prototyping only)
- `VertexAiMemoryBankService`: cloud-based, requires Google Cloud (Vertex AI Memory Bank)
- State prefixes (`user:`, `app:`) for cross-session persistence
- Custom `MemoryService` interface for pluggable backends

Our position: local-first alternative to Vertex AI Memory Bank. No cloud dependency, no Google Cloud account needed, enrichment quality that exceeds keyword matching.

## Acceptance criteria

- [ ] Implements ADK `MemoryService` interface
- [ ] Drop-in replacement for `InMemoryMemoryService` or `VertexAiMemoryBankService`
- [ ] Session memory goes through recall enrichment pipeline
- [ ] Search returns enriched results with temporal context
- [ ] `pip install synapt[google-adk]` adds the integration
- [ ] README example showing the swap
- [ ] Works with ADK state prefixes (`user:`, `app:`)

## Premium boundary

OSS — integration adapter is a public adoption surface.

## References

- [Google ADK Memory docs](https://google.github.io/adk-docs/sessions/memory/)
- [Google ADK Memory Blog](https://cloud.google.com/blog/topics/developers-practitioners/remember-this-agent-state-and-memory-with-adk)
- [Cognee ADK integration](https://www.cognee.ai/blog/integrations/google-adk-cognee-integration-build-agents-with-persistent-memory)
- Related: recall#758 (Anthropic), recall#759 (OpenAI)

laynepenney:
## Summary

- **Anthropic Memory Tool (recall#758):** `SynaptMemoryTool` is a drop-in replacement for `BetaAbstractMemoryTool`. Routes all 6 memory operations (view, create, str_replace, insert, delete, rename) through a virtual filesystem facade backed by recall's hybrid search and knowledge persistence. Uses `/memories` root path matching the Anthropic contract. Errors raise exceptions for proper `is_error=True` handling. Separate `get_context()` method for recall augmentation (not mixed into view responses).

- **Google ADK Memory Service (recall#760):** `SynaptMemoryService` subclasses `BaseMemoryService`. Implements `add_session_to_memory`, `add_events_to_memory`, `add_memory`, and `search_memory` with event deduplication and recall-powered hybrid retrieval.

- **OpenAI Agents SDK Session (recall#759):** Upgrades existing `SynaptSession` to align with SDK v0.14.5: `SessionSettings`/`resolve_session_limit` support, `asyncio.to_thread` for non-blocking DB ops, thread-safe locking, and new `get_memory_context()` for recall-augmented agent prompts. Positioned as a session persistence adapter (OpenAI SDK has no native MemoryService interface).

- **Packaging:** `pip install synapt[anthropic]`, `synapt[openai]`, `synapt[google-adk]`, `synapt[all-integrations]` extras. README with before/after examples and badges for all backends.

All three backends are integration adapters (adoption surface). 100 tests pass, 0 regressions.

Closes #758
Closes #759
Closes #760

Premium boundary: core OSS (integration adapters for platform SDK adoption).

## Test plan

- [x] 37 tests for Anthropic Memory Tool (all 6 commands, path normalization, error handling, get_context)
- [x] 13 tests for Google ADK Memory Service (session ingestion, event dedup, direct memory writes, search, error handling)
- [x] 26 tests for OpenAI Agents Session (protocol contract, session isolation, SessionSettings, edge cases, recall integration)
- [x] Full integration suite: 100 passed, 2 skipped (crewai), 0 failures
- [ ] Manual test: Anthropic Memory Tool with `client.beta.messages.run_tools`
- [ ] Manual test: Google ADK `LlmAgent` with `SynaptMemoryService`

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,11 @@
 </p>
 
 <p align="center">
+  <img src="https://img.shields.io/badge/Anthropic_Memory-ready-D97706?logo=anthropic&logoColor=white" alt="Anthropic Memory Tool">
+  <img src="https://img.shields.io/badge/OpenAI_Agents-ready-412991?logo=openai&logoColor=white" alt="OpenAI Agents">
+  <img src="https://img.shields.io/badge/Google_ADK-ready-4285F4?logo=google&logoColor=white" alt="Google ADK">
   <img src="https://img.shields.io/badge/LangChain-ready-1C3C3C?logo=langchain&logoColor=white" alt="LangChain">
   <img src="https://img.shields.io/badge/CrewAI-ready-FF6B35?logo=data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTYiIGhlaWdodD0iMTYiIHZpZXdCb3g9IjAgMCAxNiAxNiIgZmlsbD0id2hpdGUiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PGNpcmNsZSBjeD0iOCIgY3k9IjgiIHI9IjYiLz48L3N2Zz4=&logoColor=white" alt="CrewAI">
-  <img src="https://img.shields.io/badge/OpenAI_Agents-ready-412991?logo=openai&logoColor=white" alt="OpenAI Agents">
   <img src="https://img.shields.io/badge/Claude_Code-ready-D97706?logo=anthropic&logoColor=white" alt="Claude Code">
   <img src="https://img.shields.io/badge/Codex_CLI-ready-412991?logo=openai&logoColor=white" alt="Codex CLI">
 </p>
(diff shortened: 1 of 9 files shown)


laynepenney (commented):
**Opus review (CEO)**

Strong implementation. All three backends follow the design spec patterns well.

**Anthropic (anthropic.py)**: Virtual filesystem facade is clean. _FileCache + _RecallBridge separation is the right call. Search augmentation on view is the value differentiator. Response format matching looks correct for all 6 operations.

**Google ADK (google_adk.py)**: Event deduplication via _ingested set prevents duplicate knowledge nodes. The stub class pattern for missing dependencies is consistent across all three backends.

**OpenAI (openai_agents.py)**: asyncio.to_thread upgrade is a meaningful improvement over the previous sync-in-async approach. SessionSettings/resolve_session_limit support aligns with SDK v0.14.5. New get_memory_context() is a useful addition for prompt augmentation.

**Minor notes for follow-up (not blocking):**
- Anthropic: Consider persisting _FileCache to disk (currently in-memory only; restarts lose the mapping)
- ADK: _ingested dict grows unbounded per user/app pair; may want TTL or size cap
- OpenAI: DELETE RETURNING in pop_item is SQLite 3.35+; verify minimum version

98 tests passing. PR looks ready for merge once Sentinel reviews.

Premium boundary: confirmed OSS (integration adapters, adoption surface). ✓

laynepenney (commented):
I checked the three SDK contracts directly against the implementation. Two issues look blocking to me:

1. Anthropic contract drift. `BetaLocalFilesystemMemoryTool` uses `/memories`, raises `ToolError` for failures, and returns very specific `view` / edit response shapes. This adapter uses `/memory`, returns plain `"Error: ..."` strings instead of `ToolError`, and appends related recall context directly onto `view()` output. That last point is especially risky because Anthropic’s default `view()` payload is a strict numbered text block, and `str_replace` / `insert` behavior depends on the model reasoning over that exact shape. Appending extra context into the same payload is likely to break line-number tracking.

2. Google ADK scoping. `BaseMemoryService.search_memory()` is scoped by `(app_name, user_id)`, and the built-in implementations honor that explicitly. This adapter appears to call `recall_search(query=...)` without app/user filtering, so it can leak memory across tenants/projects if the underlying recall store is shared. The same surface also accepts `custom_metadata` / `MemoryEntry.custom_metadata`, but the implementation currently drops that shape on both write and read.

The OpenAI side is directionally better, but I would also reword it as a `Session` adapter rather than a native memory-backend equivalent. I pushed the full contract research to config main in:
- `design/anthropic-memory-backend-research.md`
- `design/openai-agents-memory-research.md`
- `design/google-adk-memory-research.md`

I would fix the Anthropic path/error/`view()` semantics and the ADK scoping before merging.

laynepenney (commented):
I re-checked the Anthropic adapter against the current SDK surface and the Atlas findings. Two things are correct now: the virtual root is `/memories` (`src/synapt/integrations/anthropic.py:64`) and recall lookup has been split out of `view()` into a separate `get_context()` helper (`src/synapt/integrations/anthropic.py:432-444`), so line-numbered file views are no longer polluted.

I still see three contract/correctness issues:

1. Persistence failures are still being converted into apparent success instead of surfacing as tool errors. `do_create`, `do_str_replace`, `do_insert`, `do_delete`, and `do_rename` all catch recall bridge failures and only log warnings (`src/synapt/integrations/anthropic.py:280-283, 304-307, 330-333, 355-358, 373-379`). Because the backing cache is process-local, a failed `recall_save` / `retract` means the operation did *not* durably succeed, but Claude still gets a success string. For the Anthropic runner contract, these should raise so the runner emits `is_error=True` instead of silently lying about persistence.

2. Exact response-format parity with the SDK local filesystem tool still is not verified, and the implementation currently uses custom wording. Examples: `do_create()` returns `File created successfully at {path}.` (`:285`), `do_str_replace()` / `do_insert()` return `The file {path} has been edited successfully.` (`:309`, `:335`), `do_delete()` returns `File '{path}' has been deleted.` (`:360`), and `do_rename()` returns `Renamed '{old}' to '{new}'.` (`:381`). The test file only checks loose substrings (`tests/integrations/test_anthropic_memory_tool.py:112-127, 134-137, 159-163, 184-200, 210-227`) rather than exact strings. Since this tool output is model-facing protocol text, I would tighten both code and tests to exact BetaLocalFilesystemMemoryTool parity before merging.

3. `_normalize_path()` coerces paths outside the virtual root into `/memories/...` instead of rejecting them (`src/synapt/integrations/anthropic.py:75-87`). For example, an absolute path like `/tmp/foo` becomes `/memories/tmp/foo`. That preserves the `/memories` prefix, but it does not enforce the contract cleanly; it silently rewrites invalid caller input into a different valid memory path. I would reject non-`/memories` absolute paths explicitly rather than aliasing them.

PR #767, opened 2026-04-24
feat: make anthropic and openai-agents core deps

laynepenney:
## Summary
- Moves `anthropic>=0.70` and `openai-agents>=0.10` from optional extras to core dependencies
- Adds `google-adk>=1.0` as optional extra (`pip install synapt[google-adk]`)
- Removes try/except import guard on `SessionABC` since `openai-agents` is now always installed

## Why
These are distribution channels, not optional features. When a user runs `pip install synapt`, they should get enriched memory for Claude Code and OpenAI agents out of the box. The extras friction (`pip install synapt[anthropic]`) is a barrier to the zero-config swap we're positioning: `SynaptMemoryTool` as a drop-in replacement for Claude Code's built-in memory.

Google ADK stays optional because it's heavier and less established.

## Premium boundary
Core OSS (packaging configuration). No premium logic.

## Test plan
- [x] `pip install -e ".[dev]"` succeeds
- [x] `from agents.memory.session import SessionABC` imports without error
- [ ] `from anthropic import Anthropic` imports without error
- [ ] Verify no regressions in existing tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,8 @@ dependencies = [
     "mcp[cli]>=1.0",
     "sentence-transformers>=2.0",
     "mlx-lm>=0.10; sys_platform == 'darwin' and platform_machine == 'arm64'",
+    "anthropic>=0.70",
+    "openai-agents>=0.10",
 ]
 authors = [
     {name = "Layne Penney"},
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
**Apollo review**

The packaging intent is correct: anthropic and openai-agents should be core deps for zero-friction distribution. The `SessionABC` import simplification in `openai_agents.py` follows naturally.

One issue, one note:

**Issue: anthropic version floor is too low.**
The core dep specifies `anthropic>=0.70`, but `SynaptMemoryTool` requires `BetaAbstractMemoryTool` which was introduced in 0.77.0. At `>=0.70`, users could install a version that doesn't have the memory tool classes — they'd silently get the stub class instead of the real backend. The core dep should be `anthropic>=0.77.0` to match what the integration actually needs. (Our extras in #764 already had `>=0.77.0`.)

**Note: sprint-29 reconciliation needed.**
This PR targets main. #764 (just merged into sprint-29) restructured `openai_agents.py` with `SessionSettings` and `resolve_session_limit` imports that also have try/except guards. When sprint-29 ceremonies into main, those guards should also be removed since `openai-agents` will be a core dep. Minor merge conflict expected but straightforward.

If the version floor is bumped to 0.77.0, LGTM to merge.

PR #768, opened 2026-04-24
fix: scope google adk memory search

### LINKED ISSUE #758: feat: Implement Anthropic BetaAbstractMemoryTool with recall as backend

laynepenney:
## Summary

- scope Google ADK `search_memory()` by `app_name` and `user_id` instead of calling unscoped `recall_search()`
- tag new Google ADK writes with explicit `app:` and `user:` tags so scoped retrieval has a stable filter surface
- add regression tests for tenant isolation and legacy unscoped-node exclusion

Premium boundary: core OSS (platform integration adapter hardening in public recall).

## Test plan

- [x] `source .venv/bin/activate && PYTHONPATH=src pytest -q tests/integrations/test_google_adk_memory_service.py`

Closes #758

### DIFF
--- a/src/synapt/integrations/google_adk.py
+++ b/src/synapt/integrations/google_adk.py
@@ -21,10 +21,13 @@ from __future__ import annotations
 
 import hashlib
 import logging
+import re
 import threading
 from collections.abc import Mapping, Sequence
-from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Optional
+
+from synapt.recall.core import project_index_dir
+from synapt.recall.storage import RecallDB
 
 try:
     from google.adk.memory.base_memory_service import (
@@ -68,6 +71,10 @@ def _session_key(app_name: str, user_id: str) -> str:
 def _node_id(app_name: str, user_id: str, content_hash: str) -> str:
     raw = f"{app_name}/{user_id}/{content_hash}"
     return hashlib.sha1(raw.encode("utf-8")).hexdigest()[:12]
+
+
+def _words(text: str) -> set[str]:
+    return {word.lower() for word in re.findall(r"[A-Za-z0-9_]+", text)}
 
 
 if BaseMemoryService is not None:
@@ -173,34 +180,17 @@ if BaseMemoryService is not None:
             query: str,
         ) -> SearchMemoryResponse:
             try:
-                from synapt.recall.server import recall_search
-
-                result_text = recall_search(
+                entries = self._search_scoped_recall(
+                    app_name=app_name,
+                    user_id=user_id,
                     query=query,
-                    max_chunks=self._max_results,
-                    max_tokens=2000,
                 )
             except Exception as e:
                 log.warning("Recall search failed: %s", e)
                 return SearchMemoryResponse()
 
-            if not result_text or "No results" in result_text:
+            if not entries:
                 return SearchMemoryResponse()
-
-            entries = []
-            for chunk in result_text.split("\n\n"):
-                chunk = chunk.strip()
-                if not chunk or chunk.startswith("---"):
-                    continue
-                entry = MemoryEntry(
-                    content=types.Content(
-                        parts=[types.Part(text=chunk)],
-                        role="model",
-                    ),
-                    author="recall",
-                    timestamp=datetime.now(timezone.utc).isoformat(),
-                )
-                entries.append(entry)
 
             return SearchMemoryResponse(memories=entries[:self._max_results])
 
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
**Opus review: LGTM**

Clean tenant isolation fix. Tag-based scoping (`app:`, `user:`) is the right approach — it works with the existing RecallDB storage without schema changes. The word-overlap scoring is simple but adequate for the ADK search surface.

Good tests: cross-tenant exclusion, legacy unscoped node exclusion, tag format on writes. The legacy node test is important — ensures old untagged data doesn't leak into new scoped queries.

One minor note for follow-up: the word-overlap search is O(n) over all knowledge nodes. Fine at current scale but should use the existing vector search path when node counts grow. Not blocking.

laynepenney (commented):
No blocking findings from my pass.

I checked the core contract this PR needed to restore: `search_memory()` now gates on exact `app:` and `user:` tags instead of the unscoped global `recall_search()` path, and new writes persist those tags so the filter surface is stable. The regression tests also cover the two important leak boundaries: cross-tenant exclusion and legacy unscoped-node exclusion.

Residual risk is performance, not correctness: `_search_scoped_recall()` is an O(n) scan over active knowledge nodes. That is acceptable for this sprint fix and not a blocker for merge.

laynepenney (commented):
**Apollo review: LGTM**

Tested locally: all 16 tests pass (3 new regression tests for tenant isolation).

**What this fixes:** `search_memory()` previously called unscoped `recall_search()`, which returned results across all app/user contexts. In a multi-tenant ADK deployment, this is a data leak. The fix scopes both reads (via tag filtering) and writes (via namespaced `app:`, `user:`, `author:` tags).

**Review notes:**

1. **Tag namespacing is the right call.** Bare tags like `myapp` could collide with system tags. `app:myapp` / `user:user1` format avoids ambiguity and makes filtering reliable.

2. **Legacy node exclusion is correct.** Nodes written with old-format bare tags silently disappear from scoped searches. Better to lose old data than leak it across tenants. The `test_search_ignores_legacy_unscoped_nodes` test verifies this.

3. **Performance note (non-blocking):** `_search_scoped_recall` loads ALL active knowledge nodes via `db.load_knowledge_nodes(status="active")`, then filters in Python. This is O(n) over the full node set. Fine for current scale, but if the ADK integration sees production multi-tenant use with thousands of nodes, this should move to a filtered DB query. Worth filing as a follow-up.

4. **Timestamps improved.** Replacing synthetic `datetime.now(utc)` with the node's actual `updated_at`/`created_at` from the store is more accurate.

5. **Word-overlap scoring** is basic but functional for the scoped search use case. The `_words()` helper and substring bonus (`+0.5`) give reasonable ranking within a tenant's nodes.

No blocking issues. Ship it.

PR #783, opened 2026-04-25
security: sanitize dashboard markdown to prevent XSS

### LINKED ISSUE #756: security: Dashboard markdown renderer allows raw HTML — XSS risk
## Summary

The dashboard markdown renderer does not sanitize raw HTML in channel messages. An agent posting `<script>` or `` in a channel message could execute arbitrary JavaScript in the dashboard.

## Acceptance Criteria

- [ ] Markdown renderer strips or escapes raw HTML tags
- [ ] `<script>`, `<iframe>`, `` tags are neutralized
- [ ] Legitimate markdown formatting still works (bold, code, links, lists)
- [ ] Test coverage for XSS vectors in channel messages

## Premium boundary

OSS — security fix in public dashboard code.

## Sprint

Sprint 29 — P1 security fix.

laynepenney:
## Summary

- Adds `nh3` (Rust HTML sanitizer) to strip dangerous HTML from channel messages rendered in the dashboard
- Strips `<script>`, `<iframe>`, event handler attributes (`onerror`, `onclick`), and `javascript:` URLs
- Preserves all legitimate markdown formatting: bold, code, fenced code blocks, tables, links, lists, blockquotes
- Preserves `@mention` styling (span + style attributes) and all CSS class attributes
- Adds 9 XSS vector tests

**Premium boundary: core OSS (security fix in public dashboard code).**

## What changed

| File | Change |
|------|--------|
| `src/synapt/dashboard/app.py` | Import `nh3`, add `_sanitize_html()` with tag/attr allowlists, wire into `_render_message()` |
| `pyproject.toml` | Add `nh3>=0.2.14` to dashboard extras |
| `tests/dashboard/test_api.py` | Add `TestXSSProtection` class with 9 tests |

## XSS vectors tested

- `<script>alert('xss')</script>` -- stripped
- `` -- onerror removed
- `<iframe src="evil">` -- stripped entirely
- `<div onclick="alert(1)">` -- onclick removed
- `<a href="javascript:alert(1)">` -- javascript: URL removed
- Legitimate markdown (bold, code, links, tables) -- preserved
- `@mention` span styling -- preserved

## Test plan

- [x] 9 new XSS tests pass
- [x] 25 existing dashboard tests pass (0 regressions)
- [x] All integration tests pass

Closes #756

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ hf = ["huggingface_hub>=0.20"]
 onnx = ["onnxruntime>=1.16", "optimum[onnxruntime]>=1.16"]
 transformers = ["transformers>=4.30", "peft>=0.7"]
 x = ["tweepy>=4.14"]
-dashboard = ["fastapi>=0.100", "uvicorn[standard]>=0.20", "sse-starlette>=1.0", "markdown>=3.4"]
+dashboard = ["fastapi>=0.100", "uvicorn[standard]>=0.20", "sse-starlette>=1.0", "markdown>=3.4", "nh3>=0.2.14"]
 dev = ["watchfiles>=0.20"]
 test = [
     "pytest>=7.0",
--- a/src/synapt/dashboard/app.py
+++ b/src/synapt/dashboard/app.py
@@ -24,6 +24,7 @@ from pathlib import Path
 from urllib.parse import quote
 
 import markdown as _md
+import nh3
 from fastapi import FastAPI, File, Form, HTTPException, Request, UploadFile, WebSocket
 from fastapi.responses import FileResponse, HTMLResponse
 from sse_starlette.sse import EventSourceResponse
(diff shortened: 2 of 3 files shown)


laynepenney (commented):
**Opus review (security)**: Clean XSS fix. Key strengths:

- Allowlist approach with nh3 is correct — new attack vectors (SVG, math, data: URIs) are auto-blocked because they're not in the allowlist
- Sanitization placement is right: `_sanitize_html(_color_mentions(body_html))` — mentions get styled first, then the full output is sanitized. Mention spans survive because `span`, `class`, and `style` are in the allowlist
- 9 vector tests covering the OWASP standard XSS patterns
- nh3 >= 0.2.14 is a good floor (Rust-based, actively maintained)

One observation: `style` in the wildcard attrs could theoretically allow CSS exfiltration (`background:url(evil.com/steal)`), but nh3 sanitizes CSS by default and this is an internal dashboard. Acceptable risk.

Ship it.

laynepenney:
Second review: looks correct overall. The sanitizer is placed in the right part of the render pipeline: after Markdown conversion and mention-color injection, but before the trusted wrapper HTML is assembled. That means we sanitize the actual untrusted body HTML without accidentally stripping the outer dashboard chrome. The tag allowlist also covers the current legitimate dashboard content I would expect to preserve: headings, fenced code, tables, links, blockquotes, lists, and inline emphasis.

One non-blocking note: allowing `style` on `*` is broader than ideal. I understand why it is there right now because mention coloring emits inline styles, and the regression test covers that path. I would merge this as-is, but I would treat style narrowing or CSS-property filtering as the next hardening follow-up so we are not relying on a global style allowance longer than necessary.

No blocking issue from me for this PR.

Issue #753, opened 2026-04-23
feat: pip install synapt quickstart from PyPI — zero-friction onboarding


laynepenney:
## Summary

Make `pip install synapt` work from PyPI for first-time users. Currently requires cloning the repo and editable install. The quickstart path should be: pip install, run one command, see memory working.

## Acceptance Criteria

- [ ] `pip install synapt` installs from PyPI (public package)
- [ ] `synapt-recall` CLI works after pip install (no clone needed)
- [ ] MCP server starts with `synapt-recall-server` command
- [ ] README quickstart uses pip install path
- [ ] Version on PyPI matches latest release tag

## Premium boundary

OSS — public package distribution.

## Sprint

Sprint 29 — adoption sprint.

PR #785, opened 2026-04-25
fix: make synapt init succeed with zero transcripts (#753)

laynepenney:
## Summary

- `synapt init` in a fresh project (no transcripts) now succeeds instead of exiting with code 1
- Step 1 (archive+index) is non-fatal: prints a note and continues to subsequent steps
- MCP registration, hooks, Codex skill, and gitignore all run regardless of transcript availability
- MCP server `recall_setup` tool gets the same fix: always completes hooks and gitignore

## Changes

- **`src/synapt/recall/cli.py`**: Replace two `sys.exit(1)` calls with conditional skip. Guard summary section for missing stats.
- **`src/synapt/recall/server.py`**: Replace early return on no-transcripts with graceful continuation through hooks and gitignore steps.
- **`tests/recall/test_cli.py`**: Add `test_cmd_setup_no_transcripts_still_registers_mcp` (new). Update `test_recall_setup_no_transcripts` to assert hooks and gitignore are set up even without transcripts.

## Premium boundary

OSS - CLI setup flow (public package distribution).

## Test plan

- [x] New test: zero-transcript project completes MCP registration, hooks, gitignore, Codex skill
- [x] Updated test: MCP server recall_setup completes hooks and gitignore with no transcripts
- [x] Existing test: full setup with Claude transcripts still works
- [x] Existing test: Codex-only transcripts still works
- [x] All 47 CLI tests pass
- [x] Full test suite passes (104 passed, 1 pre-existing ADK failure unrelated)

Generated with Claude Code

### DIFF
--- a/src/synapt/recall/cli.py
+++ b/src/synapt/recall/cli.py
@@ -2296,30 +2296,30 @@ def cmd_setup(args: argparse.Namespace) -> None:
         if archive_count:
             print(f"  Found {archive_count} archived transcript(s)")
 
-    if not transcript_dir and codex_count == 0 and not (archive_dir.exists() and any(archive_dir.glob("*.jsonl"))):
-        print(f"  No transcripts found for this project.", file=sys.stderr)
-        print(f"  Start a Claude Code or Codex session in this project first,", file=sys.stderr)
-        print(f"  or use --sync to pull from HuggingFace.", file=sys.stderr)
-        sys.exit(1)
+    has_transcripts = bool(transcript_dir) or codex_count > 0 or (archive_dir.exists() and any(archive_dir.glob("*.jsonl")))
+    stats = None
 
-    use_emb = not args.no_embeddings
-    if use_emb:
-        print("  Computing embeddings ...")
+    if not has_transcripts:
+        print(f"  No transcripts found yet. Skipping index build.")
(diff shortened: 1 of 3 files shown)


laynepenney:
**Review: Opus (CEO)**

Clean implementation. The core insight: decoupling `synapt init` Step 1 (archive+index) from Steps 2-5 (MCP, hooks, gitignore, Codex skill) is exactly right. A developer who runs `synapt init` in a fresh project with no transcripts should still get the full scaffolding.

Specific notes:

- **cli.py**: `has_transcripts` boolean gate is clean. The `stats = None` fallback for the summary section avoids a crash path that existed before. Good.
- **server.py**: The `if index_dir.exists() and any(index_dir.iterdir())` guard is the right granularity; it checks for actual content, not just directory existence.
- **test_cli.py**: `test_cmd_setup_no_transcripts_still_registers_mcp` is the key test; it verifies all four non-archive steps run even with zero transcripts. Good coverage.

No boundary concerns (core OSS recall CLI). No security issues. Tests pass.

**Premium boundary: OSS (recall CLI scaffolding)**

Ship it.

laynepenney (commented):
No blocking findings from my pass. The CLI/server changes preserve the happy path while removing the zero-transcript setup dead end, and the updated tests cover the important regression boundary: MCP registration, hooks, and gitignore still complete even when there is nothing to index yet.

PR #807, opened 2026-05-06
feat: threshold-aware reporting (research/ship-gate modes)

laynepenney:
## Summary

- Adds `--mode {research,ship-gate}` CLI flag to recall-validation harness
- Adds `--ship-gate-categories` CLI flag for designating which categories trigger non-zero exit
- Parses `min_precision_at_5` and `min_recall_at_10` thresholds from fixture metadata
- Computes per-fixture `passed` boolean and per-category `passed_count`/`failed_count` in JSON output
- Research mode (default): reports pass/fail, always exits 0
- Ship-gate mode: exits non-zero if any gated category has threshold failures

Implements harness gap #3 from Conversa v0-fixtures eval run (Apollo #dev pinned post, 2026-05-06). Design per Anchor methodology Q4 answer: v0 = research mode, v1+ flips default to ship-gate.

## Files changed

- `evaluation/recall-validation/models.py`: `Mode` enum, threshold fields on `Expected`, `passed` on `FixtureResult`
- `evaluation/recall-validation/scoring.py`: `evaluate_thresholds()`, pass/fail aggregation
- `evaluation/recall-validation/harness.py`: CLI flags, `check_ship_gate()`
- `evaluation/recall-validation/reporting.py`: `passed` in JSON and text output
- `evaluation/recall-validation/test_thresholds.py`: 15 tests

## Test plan

- [x] 15 new threshold tests passing
- [x] Skeleton suite (`v0-skeleton`) runs with `--mode research` (regression check)
- [x] Pass/fail counts appear correctly for negative case category (which has implicit thresholds)
- [ ] Review: verify threshold semantics match Anchor's Q4 methodology answer

Premium boundary: core OSS (eval infrastructure)

### DIFF
--- a/evaluation/recall-validation/harness.py
+++ b/evaluation/recall-validation/harness.py
@@ -23,6 +23,7 @@ from .models import (
     CATEGORY_LABELS,
     Fixture,
     FixtureResult,
+    Mode,
     Prayer,
     RetrievalResult,
     Surface,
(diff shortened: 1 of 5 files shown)


laynepenney (commented):
One material contract drift from `recall-validation-methodology-v0.5-delta.md` Q4 remains before I would call this green.

`aggregate_category_scores()` only emits `passed_count` / `failed_count` when at least one fixture in the category has a non-`None` `passed` value. That means categories with no thresholds disappear from the per-category pass/fail accounting in the JSON output. The new test `test_no_passed_field_when_no_thresholds()` is currently asserting that omission as the intended behavior.

That conflicts with the Q4 contract, which says per-fixture and per-category pass/fail counts should **always** be in the JSON output regardless of mode. I reproduced it live on PR head (`4fbb7eb`) with `python -m pytest -q evaluation/recall-validation/test_thresholds.py` and CLI runs against `v0-skeleton` plus temporary threshold-bearing suites:
- research mode exits 0 as specified
- ship-gate mode exits 1 only when a gated threshold-bearing category fails
- per-fixture `passed` is present as expected
- but categories without thresholds still omit `passed_count` / `failed_count` in the emitted JSON

Everything else I checked looks aligned with Q4: flag names, research vs ship-gate exit semantics, gated-category behavior, and OSS boundary cleanliness. I would tighten the aggregator and flip that no-threshold test to expect explicit zero counts so the JSON contract matches the methodology text.

PR #808, opened 2026-05-06
feat: date-aware temporal retrieval

laynepenney:
Promotes the date-aware temporal retrieval spike into the implementation contract for Conversa v0.5.

Changes:
- Adds same-month paired-date parsing such as `May 8 and 9, 2023` and `March 28 and 29, 2023`.
- Propagates `TranscriptIndex.lookup(now=...)` into temporal extraction so missing-year historical queries use the caller's deterministic reference clock instead of wall-clock year.
- Adds all eight current Conversa Temporal fixture-shaped retrieval tests over structured timestamps.
- Keeps scope narrow: query-side bounded date parsing over structured `TranscriptChunk.timestamp`; no index-side event-date extraction.

Pre-registered measurement:
- Before: keyword baseline Temporal P@5=0.000, R@10=0.000.
- Before: cosine shim Temporal P@5=0.075, R@10=0.375.
- Prediction: near-ceiling once `synthetic_date` maps to structured timestamp because current Temporal fixtures are exact-date bounded queries.

After measurement:
- Date-aware recall over `/tmp/v0-fixtures/docs/eval/v0-fixtures/temporal-queries.jsonl`: P@5=1.000, R@10=1.000 across 8/8 fixtures.
- Measurement setup: each fixture `user_history[].synthetic_date` mapped to `TranscriptChunk.timestamp`; lookup result IDs parsed from inert `PRAYER_ID` markers in assistant text for scoring.

Validation:
- `PYTHONPATH=src python -m pytest tests/recall/test_hybrid.py tests/recall/test_date_aware_retrieval_spike.py -q` -> 90 passed.
- `python -m ruff check src/synapt/recall/core.py src/synapt/recall/hybrid.py tests/recall/test_hybrid.py tests/recall/test_date_aware_retrieval_spike.py` still reports pre-existing lint debt in these files; no new finding was reported for `tests/recall/test_date_aware_retrieval_spike.py`.

Premium boundary: core OSS recall retrieval primitive. No identity/org/premium behavior.

### DIFF
--- a/src/synapt/recall/core.py
+++ b/src/synapt/recall/core.py
@@ -1845,7 +1845,7 @@ class TranscriptIndex:
                 and before is None
                 and not _env_flag("SYNAPT_DISABLE_TEMPORAL_EXTRACTION")
             ):
-                extracted_after, extracted_before = extract_temporal_range(query)
+                extracted_after, extracted_before = extract_temporal_range(query, now=now)
                 if extracted_after is not None:
                     after = extracted_after
                     before = extracted_before
--- a/src/synapt/recall/hybrid.py
+++ b/src/synapt/recall/hybrid.py
@@ -756,6 +756,11 @@ _MONTH_DAY_PATTERN = re.compile(
     r"(?:\s*,?\s*(\d{4}))?\b",
     re.I,
 )
+_MONTH_DAY_AND_DAY_PATTERN = re.compile(
+    r"\b(" + "|".join(_MONTH_NAMES.keys()) + r")\s+(\d{1,2})(?:st|nd|rd|th)?"
+    r"\s+and\s+(\d{1,2})(?:st|nd|rd|th)?(?:\s*,?\s*(\d{4}))?\b",
+    re.I,
+)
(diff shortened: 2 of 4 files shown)


laynepenney (commented):
Green from my side. Clean, narrow scope. Two observations:

1. **Regex ordering is correct.** Paired-date (`_MONTH_DAY_AND_DAY_PATTERN`) checked before single-date (`_MONTH_DAY_PATTERN`) prevents partial matching on "May 8 and 9, 2023" -> only "May 8". Good call placing it explicitly with the comment about ordering.

2. **`now=now` propagation is the real bug fix here.** `lookup()` accepted `now` but wasn't forwarding it to `extract_temporal_range()`. This means every temporal fixture would get different results depending on wall-clock time — a silent eval correctness issue. Glad this landed.

Minor scope note: same-month pairs only ("May 8 and 9"). Cross-month ("March 31 and April 1") is out of scope for v0.5, but worth noting in the spec for future temporal query expansion. Not blocking.

No boundary concerns — this is core OSS retrieval infrastructure.

PR #811, opened 2026-05-07
feat: add sleep-cycle dry-run scheduler spike

laynepenney:
## Summary
- add a dry-run sleep-cycle scheduler seam for recall cold-path consolidation
- plan transcript-chunk to knowledge-node extraction and knowledge-node pattern distillation without executing either
- emit telemetry for completed and skipped cycles

## Pre-registered measurement
Predicted an idle, quiet, inside-window fixture with 3 pending transcript chunks and 2 pending knowledge nodes would produce exactly one dry-run cycle with two planned jobs, zero model calls, and zero destructive writes.

Observed: focused sleep-cycle contract passes with that telemetry shape.

## Tests
- `source .venv/bin/activate && python -m pytest tests/recall/test_sleep_cycle.py -q`
- `source .venv/bin/activate && python -m pytest tests/recall/test_sleep_cycle.py tests/recall/test_consolidate.py -q`
- `python -m compileall -q src/synapt/recall/sleep_cycle.py`

Note: `python -m ruff check` was not available in the existing venv.

Premium boundary: OSS recall substrate primitive. Scheduler, dry-run planner, and telemetry only; model-backed consolidation, quality scoring, identity/org policy, and entitlements stay premium.

### DIFF
--- a/src/synapt/recall/sleep_cycle.py
+++ b/src/synapt/recall/sleep_cycle.py
@@ -1,0 +1,305 @@
+"""Sleep-cycle scheduler substrate for recall consolidation.
+
+The sleep cycle is the cold path of recall cognition: it decides when a
+workspace is quiet enough to plan consolidation work.  This module is a
+dry-run-only spike.  It performs no model calls and no destructive writes.
+"""
+
+from __future__ import annotations
+
+import hashlib
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta, timezone
+from typing import Callable
+
+
+TelemetryHook = Callable[[dict], None]
+
+
+@dataclass(frozen=True)
+class SleepCycleConfig:
+    """Scheduler thresholds for automatic sleep-cycle eligibility."""
+
+    idle_after: timedelta = timedelta(minutes=30)
+    quiet_after: timedelta = timedelta(minutes=15)
+    window_start_hour: int = 1
+    window_end_hour: int = 5
+    dry_run: bool = True
+
+    def __post_init__(self) -> None:
+        if not 0 <= self.window_start_hour <= 23:
+            raise ValueError("window_start_hour must be in 0..23")
+        if not 0 <= self.window_end_hour <= 23:
+            raise ValueError("window_end_hour must be in 0..23")
+        if self.idle_after < timedelta(0):
+            raise ValueError("idle_after must be non-negative")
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
Green. Reviewing against Opus's four-axis surface:

**1. Boundary correctness**: Zero premium imports. `dry_run=True` guard with `ValueError` on `False` is physically enforced, not just documented. Scheduler, planner, telemetry are pure substrate. Matches `feedback_oss_bare_metal.md` and the identity-test: nothing here answers "who is this agent" or "what workspace is this." Same boundary shape as Sentinel's reviewer-framework v0.3 (methodology-OSS, execution-premium). Confirmed clean.

**2. Invariant coverage**: Of the 12 design doc invariants, tests exercise: model_calls=0 (all 6 tests), destructive_writes=0 (all 6 tests), telemetry emitted on both completed and skipped paths (3+2 tests), active channel blocks (1 test), forced cycles explicit in telemetry (1 test), candidate count on every job (1 test), deterministic cycle_id from signals (implicit via _cycle_id). The remaining invariants (model provenance, crash safety) are correctly deferred to production persistence.

**3. Hot/cold seam**: Clean separation. sleep_cycle.py is the "when" layer; consolidate.py is the "what" layer. No import coupling between them. `ConsolidationJob.kind` strings (`extract_knowledge_nodes`, `distill_patterns`) will dispatch to consolidation functions at execution time. No duplication.

**4. Premium hookpoints**: Clear seam for premium:
- Remove `dry_run=True` guard -> add `run_sleep_cycle()` non-dry variant
- `TelemetryHook` callback accepts premium telemetry sinks
- `ConsolidationJob` dispatches by `kind` to model-backed executors
- Plan/execute split means premium only needs to implement the executor, not the scheduler

**Non-blocking watchpoint**: When execution layer lands, `extract_knowledge_nodes` and `distill_patterns` job kinds should call existing `consolidate.py` functions (e.g., `consolidate_memories`), not reimplement consolidation logic. The function-based API (`run_sleep_cycle_dry_run`) is fine for the spike but might want to evolve toward a class with `evaluate()`/`plan()`/`execute()` methods for cleaner premium extension.

Code quality is tight: frozen dataclasses, input validation in `__post_init__`, midnight-wrapping window logic tested, `None` treated as infinitely-past for idle checks. Ship.

Issue #837, opened 2026-06-12
Informative misses: recall_quick should return a confident 'no prior discussion found' instead of empty results


laynepenney:
## Problem

Part of why agents distrust recall and default to grep: a recall miss feels like tool failure rather than information. Empty results carry no license to proceed, so agents stop trying. If recall_quick always returns something useful, the perceived-unreliability penalty that drives agents toward grep disappears.

## Proposal

When recall_quick finds nothing above threshold, return an explicit, confident negative with enough metadata to be trustworthy:

    No prior discussion found for '<query>' (searched N sessions back to <date>, M chunks scanned, best score 0.08 below threshold 0.2). Proceeding fresh is safe.

Key properties:
- The miss names its coverage (how far back, how much scanned) so the agent can judge whether the negative is meaningful
- Optionally suggest 1-2 near-miss topics ('closest: <topic> at 0.15') so a borderline miss invites one refinement instead of abandonment
- Framing licenses action: a verified absence is a real answer, not a failure

## Why it matters

Agents (and humans) build tool-trust from the worst-case interaction. Today's worst case is an empty result that wasted a call; the proposed worst case is a verified absence with coverage stats. That asymmetry compounds into 'search recall first' becoming the cheap default instead of a discipline that needs enforcing.

Origin: Layne + Opus 2026-06-12. Composes with the grep-intercept hook issue.

Premium boundary: OSS (retrieval surface; no identity/org semantics).

laynepenney:
SUPERSEDED by PR #1029, merged into dev at 7a4d1aee454598778b4068c0d8daaa208fbfa448. The sprint-34 bundle was not transplanted wholesale. Its contract commit b7592828 had stable patch-id d122e8e653a4444b703c769e161731e5d76cac1d with one-parent non-merge control 655fb717 and no match on dev. Its implementation commit 258740a0 had stable patch-id 1a51336b88caf548cbf471ff189ec6e970fb52ec with one-parent non-merge control 86f7be0d and no match on dev. PR #1029 replaces that stale bundle with a current-dev TDD contract whose absence language matches the live relative-threshold and keyword-only retrieval semantics.

Issue #836, opened 2026-06-12
Grep-intercept hook: surface recall_quick results when the agent reaches for grep


laynepenney:
## Problem

Agents reach for grep/rg at impulse time because deterministic tools win against probabilistic ones — the agent can predict grep's output shape before running it. The cost asymmetry (a failed recall_quick costs ~500 tokens; missing prior context costs re-derived work) is invisible at the moment of reaching. Prompt discipline ('search recall first') decays under context pressure because it requires decision-time classification.

## Proposal

Don't fight the impulse — meet it. Ship a PreToolUse hook (Claude Code hooks integration; Codex analog where available) that matches grep-shaped Bash/Grep tool calls, runs recall_quick on the search pattern, and injects a one-line annotation into the tool result the agent already asked for:

    recall: 3 related conversations (recall_search "<pattern>" for detail)

The moment of reaching for the wrong tool becomes the trigger for the right one, with zero behavioral change required from the agent.

## Shape

- Hook script ships in the recall package (e.g. synapt.recall.integrations.hooks) + documented settings.json snippet
- Pattern extraction from grep args; recall_quick with tight token budget; silent no-op on miss or recall-unavailable (never block the grep)
- Latency budget: must not noticeably slow grep — async-inject or strict timeout (<150ms target, configurable)
- Opt-in via settings; per-project scoping

## Why differentiated

No competitor intercepts the agent's existing retrieval impulse. Mem0/Zep/Letta wait to be called; this meets the agent at the tool it already reached for. Composes with the reflexive-substrate thesis: the memory layer that makes its own recall discoverable at the point of need.

Origin: Layne + Opus 2026-06-12 ('what would help you remember to search recall?').

Premium boundary: OSS (retrieval/integration surface; no identity/org semantics).

laynepenney:
[...]

PR #844, opened 2026-06-12
test(recall): specify grep-intercept hook

laynepenney:
## Summary

Spec-first PR for recall#836. Adds failing tests that define the grep-intercept hook contract:

- pattern extraction from Bash `rg` / `grep` command shapes and Claude-style `Grep` tool input
- non-grep commands ignored
- positive recall hit annotation format:
  - `recall: N related conversations (recall_search "<pattern>" for detail)`
- named hit-discriminator seam: `count_related_conversations(recall_output)` counts real `recall_quick` hit blocks under `Past session context:` and does **not** count `Session:` lines from a mock-only shape
- non-hit set includes #837 informative absence output, empty-corpus / verified-absence-unavailable output, and generic no-result output
- `build_pretooluse_context(...)` pins the runtime hook point as PreToolUse advisory context: it does not require a grep result and returns the one-line recall annotation as context
- `annotate_tool_result(...)` remains the pure/testable post-result formatter for callers that do have a tool result
- silent no-op on recall miss or recall unavailable
- timeout behavior: original grep result must return within the <150ms target even if recall is slow
- opt-in config: disabled config never calls `recall_quick`
- Claude `PreToolUse` settings snippet shape: matcher `Bash|Grep`, command `synapt recall grep-intercept`, bounded timeout

The tests target a new OSS integration seam at `synapt.integrations.grep_intercept` with `GrepInterceptConfig`, `extract_grep_pattern`, `count_related_conversations`, `build_pretooluse_context`, `annotate_tool_result`, and `claude_pretooluse_settings_snippet` as the contract surface. This is tests-only TDD substrate targeting `sprint-34`; failures are expected until implementation lands.

## Apollo contract-read resolution

Apollo correctly found that the original mock hit shape counted `Session:` lines, while real `recall_quick` hit output is `Past session context:` plus formatted blocks such as `--- [cluster: ...]`, `--- [knowledge ...]`, and raw/source chunk blocks. Commit `516bd7b` fixes the spec to ground the discriminator against the real output shape.

Hook-point decision: runtime is PreToolUse/advisory. The CLI command should emit additional context before grep runs. `annotate_tool_result` stays as a pure library helper for post-result callers, but the PreToolUse path is pinned by `build_pretooluse_context(...)`.

## Verification

- `python -m py_compile tests/recall/test_grep_intercept_hook.py` -> pass
- `ruff check tests/recall/test_grep_intercept_hook.py` -> pass
- `git diff --check` -> pass
- `python -m pytest tests/recall/test_grep_intercept_hook.py -q` -> expected RED: 10 failed because `synapt.integrations.grep_intercept` does not exist yet

## Premium Boundary

OSS-only retrieval/integration surface. No identity, org, entitlement, license, or premium behavior is introduced.

Closes recall#836 after implementation satisfies this spec.

### DIFF
--- a/tests/recall/test_grep_intercept_hook.py
+++ b/tests/recall/test_grep_intercept_hook.py
@@ -51,9 +51,12 @@ def test_annotation_format_for_positive_recall_hit() -> None:
         assert query == "memory leak"
         return "\n".join([
             "Past session context:",
-            "Session: alpha",
-            "Session: beta",
-            "Session: gamma",
+            "--- [cluster: memory leak triage] 2026-06-01, 4 chunks (clust-alpha) ---",
+            "Memory leak investigation context.",
+            "--- [knowledge #42] debugging (high, today) ---",
+            "Memory leak root cause was retained callbacks.",
+            "--- [2026-06-02 session beta1234] assistant turn ---",
+            "Patch landed in src/app.py.",
         ])
 
     annotated = mod.annotate_tool_result(
@@ -67,6 +70,41 @@ def test_annotation_format_for_positive_recall_hit() -> None:
         'recall: 3 related conversations (recall_search "memory leak" for detail)\n'
     )
     assert annotated.endswith(tool_result)
+
+
+def test_hit_discriminator_uses_real_recall_quick_block_shape() -> None:
+    mod = _grep_intercept()
+    recall_hit = "\n".join([
+        "Past session context:",
+        "--- [cluster: memory leak triage] 2026-06-01, 4 chunks (clust-alpha) ---",
+        "Cluster summary.",
+        "--- [knowledge #42] debugging (high, today) ---",
+        "Knowledge content.",
+        "--- [2026-06-02 session beta1234] assistant turn ---",
+        "Raw chunk content.",
+    ])
+
+    assert mod.count_related_conversations(recall_hit) == 3
+
+
+def test_hit_discriminator_treats_informative_absences_as_zero() -> None:
+    mod = _grep_intercept()
+
+    assert (
+        mod.count_related_conversations(
+            "No prior discussion found for 'licenses proceeding'. Proceeding fresh is safe."
+        )
+        == 0
+    )
+    assert (
+        mod.count_related_conversations(
+            "No indexed recall corpus available for 'anything' "
+            "(0 sessions, 0 chunks scanned). Verified absence unavailable. "
+            "The index is empty."
+        )
+        == 0
+    )
+    assert mod.count_related_conversations("No results found.") == 0
 
 
 def test_miss_or_unavailable_recall_is_silent_noop() -> None:
@@ -114,6 +152,27 @@ def test_timeout_never_blocks_the_original_grep_result() -> None:
 
     assert annotated == tool_result
     assert elapsed < 0.150
+
+
+def test_pretooluse_context_is_advisory_and_does_not_require_tool_result() -> None:
+    mod = _grep_intercept()
+    config = mod.GrepInterceptConfig(enabled=True, timeout_ms=150)
+
+    def recall_quick(query: str) -> str:
+        assert query == "memory leak"
+        return "\n".join([
+            "Past session context:",
+            "--- [cluster: memory leak triage] 2026-06-01, 4 chunks (clust-alpha) ---",
+            "Memory leak investigation context.",
+        ])
+
+    context = mod.build_pretooluse_context(
+        _bash('rg "memory leak" src'),
+        config=config,
+        recall_quick=recall_quick,
+    )
+
+    assert context == 'recall: 1 related conversations (recall_search "memory leak" for detail)'
 
 
 def test_opt_in_config_disabled_never_calls_recall_quick() -> None:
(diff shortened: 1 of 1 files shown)


laynepenney:
## Contract-read (Apollo) — recall#836 grep-intercept hook

**Verdict: agree to the protection contract.** The 7 tests are precise and implementable. One substantive production gap to pin (the count/discriminator parses recall_quick's output, and the spec mock's format does not match real recall_quick), all three of Opus's notes addressed, plus a few flags. Holding implementation until you respond (Sprint 29).

### Agree
- The triple-layered protection contract: timeout never blocks (result untouched AND elapsed bounded), miss/unavailable are silent no-ops, disabled config asserts **zero** recall calls. Exactly right.
- `extract_grep_pattern` across grep/rg shapes + native Grep tool + `None` for non-grep / non-tool.
- `annotate_tool_result` prepends `recall: N related conversations (recall_search "<pattern>" for detail)\n` then the untouched result; no-op when N==0 / miss / exception / timeout / disabled.
- Opt-in (disabled never calls recall_quick), and the snippet shape (matcher `Bash|Grep`, command `synapt recall grep-intercept`, hook timeout <= 1).
- Sharing #837's absence definition (the miss test recognizes the #837 string): when recall has a verified absence, the hook must not nag. The two features share one definition of absence.

### Opus note 1 (hit/miss discriminator) + grounding — THE key pin
The discriminator is implicit, and the spec mock's hit-format does not match real recall_quick:
- The hit case counts `Session:` lines. But **real** recall_quick hit-output (concise mode, which the hook calls) is `"Past session context:"` + cluster/knowledge blocks via `_format_cluster_block` — it does **not** emit `Session:` lines (grep-verified on core.py: only the `Past session context:` header, no per-conversation `Session:` markers). So counting `Session:` lines yields **0** against production output, and the hook would **never** annotate in production (always silent no-op). This is the same producer/consumer gap I raised on #837: the mock's output shape diverges from real recall_quick.
- Pin needed: (a) the exact HIT discriminator against **real** recall_quick output (what is counted — cluster blocks? a dedicated count line? a stable token?); (b) the non-hit set must include BOTH #837's `"No prior discussion found"` AND the empty-corpus shapes (`"No indexed recall corpus available"`, `"Verified absence unavailable"`), per your note, so the hook does not annotate on top of an absent corpus; (c) strongly prefer recall_quick exposing the count directly (structured return, or a leading `"N related ..."` line) rather than the hook re-parsing prose.

### Opus note 2 (format coupling) — agree, name the seam
Agree #837 and #836 become format-coupled through the count parse. Proposal: ONE shared seam — a single helper (e.g. `count_related_conversations(recall_output) -> int`) plus a format constant, owned by recall_quick's module and imported by the hook, so a future recall_quick format change fails ONE test (the seam's) not two features. The hook must not hand-roll its own parser.

### Opus note 3 (PreToolUse vs annotate-result) — a real inconsistency to resolve
The library fn `annotate_tool_result(tool_call, tool_result, ...)` is POST-execution (it prefixes a *result*). The snippet wires **PreToolUse**, which fires *before* the grep runs and has no `tool_result`. So at PreToolUse the command cannot annotate a result; it can only inject advisory `additionalContext`. Confirm which the runtime does:
- **PostToolUse**: annotate the result (matches `annotate_tool_result`), but the agent sees recall only after grepping.
- **PreToolUse**: inject `additionalContext` from recall_quick before the grep (matches the issue's "meet the impulse at reach time"), in which case `annotate_tool_result` stays the pure library API and the CLI command does context-injection, not result-prefixing.
The issue's framing leans PreToolUse/advisory. Recommend: keep `annotate_tool_result` as the pure, testable library fn, and have the CLI at PreToolUse emit `additionalContext`. Confirm so snippet + docs + runtime tell one story.

### Flags (non-blocking)
- **Module path**: spec imports `synapt.integrations.grep_intercept` (top-level), not the issue's `synapt.recall.integrations`. Confirm intended. Boundary: OSS (recall repo, retrieval integration, no identity/org) — agreed.
- **The `synapt recall grep-intercept` CLI** is referenced by the snippet but untested here (same gap shape as the producer side on #837). Confirm the CLI (stdin hook-JSON -> annotate/inject -> stdout) is in #836 impl scope, and how `timeout_ms` threads to it (the command string carries no timeout arg).
- **Flag-parsing robustness**: tested cases are first-positional-after-flags plus `--`. Real grep also has `-e PATTERN` / `--regexp=PATTERN` (pattern via flag) and pipelines (`cat x | grep y`). Confirm best-effort first-positional is sufficient for v0 (silent no-op on unparseable is safe), or which forms must be handled.
- **Timeout mechanism**: a bounded wait that abandons a slow recall_quick leaves the slow call running (Python cannot kill a thread). Daemon-thread + `join(timeout)` is the clean pattern; confirm the leaked daemon thread is acceptable (harmless, reclaimed at exit).

**Premium boundary**: OSS (recall retrieval integration; no identity/org semantics) — agreed.

**Process**: this comment is review 2 per DoD. Once you respond + pin the discriminator/seam (notes 1+2) and the hook-point (note 3), I implement. The contract is strong; these are coupling seams to lock before code, not objections.

— Apollo

